PR #15, opened 2015-08-25
Support for converting to typed slices.

tcolar:
This allows converting generic slices ([]interface{}) into typed slices.
Makes it easier to call Go code that takes in typed arrays / varargs.


mattn:
Sorry about delay. Thanks.

PR #20, opened 2016-03-20
A mechanism to interrupt the execution.

xrash:
This is a proposal of a mechanism to interrupt the execution of a statements chain.

There are other ways to implement it -- for instance, using an interrupt channel --, but using a simple boolean check seems to have less of a performance burden.

A consequence is that the environment can get messy after an interrupt, but I am not fully aware of the internals, and so I ask for your review and feedback in this regard.

Also, calling Interrupt() in an environment which is not running causes the next Run() to fail. To fix that, it would be nice to, maybe in another PR, build a better API for the embedded functionality. I feel like RunSingleStmt should be an internal function and (a more user-friendly version of) Run should be the only entry point of execution (for embedders).

### DIFF
--- a/vm/env.go
+++ b/vm/env.go
@@ -9,46 +9,55 @@ import (
 // Env provides interface to run VM. This mean function scope and blocked-scope.
 // If stack go to blocked-scope, it will make new Env.
 type Env struct {
-	name   string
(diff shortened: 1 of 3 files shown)


mattn:
thanks!

PR #22, opened 2016-03-22
fix undefined symbol 'string' in examples

TrueFurby:
Convertor func **string** [was replaced]( https://github.com/mattn/anko/commit/d9c438af3e18ddedc9d9d68d564900fbf2ce5bfe#diff-32b04d93ce30b8591f4ad2906b972927L141) by **toString**.

However affected examples need to display _[]byte_ as _string_ and **toString** would display it as _[]byte_, thus using printf("%s", ...) fixes this.


mattn:
Thank you

PR #23, opened 2016-03-23
Refactor API usage

TrueFurby:
I have discovered this project just a while ago when looking for some embeddable language for Go and  choosing between several options and deciding for **Anko** because I preferably wanted one that was similar to Go and active.

After playing around with it I started to get annoyed by the unnecessary redundancy of the API design, especially converting all values with reflect.ValueOf(). So I've dived into the source to see how could improve it to write less code to get same result. 

The easiest part was getting rid of reflect.Value, since it takes interface{} as parameter so the conversion can simply be hidden from API user, making the code shorter and more readable. 
To prevent breaking API for the current users I added a check to see if the underlying value of interface{} is reflect.Value and if it isn't use reflect.ValueOf() to get it.

So, future usage can simply be changed to this:
```Go
env.Define("foo", 1)

/* INSTEAD OF:
env.Define("foo", reflect.ValueOf(1)) */
```
I also added two helper functions `parser.ParseSrc(src string)` and `env.Execute(src string)`  to provide quick ways to parse and run the source code without needing to deal with Scanner or Run function. 

```Go
stmts, err := parser.ParseSrc(code)
if err != nil {
	log.Fatal(err)
}
v, err = vm.Run(stmts, env)

// OR EVEN SHORTER:
v, err := env.Execute(`foo + 1`)

/* INSTEAD OF:
scanner := new(parser.Scanner)
scanner.Init(`foo + 1`)
stmts, err := parser.Parse(scanner)
if err != nil {
	log.Fatal(err)
}
v, err := vm.Run(stmts, env) */
```


mattn:
Thanks. Need to convert builtin/**/*.go.

TrueFurby:
Done.

mattn:
Merged. https://github.com/mattn/anko/commits/master

Thanks. :+1:

Issue #30, opened 2016-05-04
Bugs in toInt, toFloat, toBool functions?


mmindenhall:
I have created a function in go that returns a uint32.  I then call toInt on that value in anko, which converts it to 0.  I think there's a bug in how this is currently defined:

```go
env.Define("toInt", func(v interface{}) int64 {
		nt := reflect.TypeOf(1)
		rv := reflect.ValueOf(v)
		if rv.Type().ConvertibleTo(nt) {
			return 0
		}
		return rv.Convert(nt).Int()
	})
```
For this to work, I think the return statements will need to switch positions ([see in playground](https://play.golang.org/p/oT56Lk3-3f), with added ability to handle converting strings).

BTW, `toFloat` and `toBool` also have this bug.  I'll submit a PR to fix all 3 (and add the capability to convert from string).

mattn:
Do you want to convert string into int or uint32? Then use strconv package.

mmindenhall:
The golang way is to use strconv.  Given that anko is less strict, I thought it would be a nice feature for `toInt` to handle converting strings as well as numeric types.  If you don't want to add that feature, no problem.  

The real problem here is that the return statements need to swap positions for this to work properly:

```go
	// this returns 0 if the type is convertible, and tries to convert if it's not!
	if rv.Type().ConvertibleTo(nt)
		return 0
	}
	return rv.Convert(nt).Int()

	// this works
	if rv.Type().ConvertibleTo(nt) {
		return rv.Convert(nt).Int()
	}
	return 0
```

As said above, `toFloat` and `toBool` also need to have return statements swapped in a similar manner.

mattn:
Ah, you are right. will fix in later.

PR #32, opened 2016-05-05
More dynamic support for toInt, toFloat, toBool

mmindenhall:
This is along the lines of what I was talking about in #30.  Run the new script to see it in action.  No worries if this isn't a feature you want to add. 
- toInt converts any string that can be parsed as int64 or float64
- toInt converts bool `true` to 1, `false` to 0.
- toFloat converts any string that can be parsed as float64
- toFloat converts bool `true` to 1.0, `false` to 0.0.
- toBool converts any positive number `true`, 0 or negative to `false`.
- toBool converts (ignoring case) "1", "t", "true", "y", "yes" to `true`, any other string to `false`

### DIFF
--- a/_example/scripts/toType.ank
+++ b/_example/scripts/toType.ank
@@ -1,0 +1,122 @@
+#!anko
+
+# toInt with ints, floats, strings, bools
+println("\ntoInt examples:\n===============")
+i = 1<<63 - 1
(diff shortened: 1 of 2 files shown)


mattn:
merged. https://github.com/mattn/anko/commit/d7753b9eec347733b7ef6a9698a4eb85531f3ace

thanks.

PR #37, opened 2016-08-12
made it easy to load all builtins for the embedded use case

dhiltonp:
I was testing anko for an embedded application, but had some issues getting examples to work.

I dug a bit (not too hard at this stage), and found that the built-ins aren't all loaded by default. I think this makes sense, but for evaluation purposes I'd like to be able to easily pull in all builtins; the exact set of desired functions can later be defined.

I factored out the builtin import code to core, and documented the change.

### DIFF
--- a/README.md
+++ b/README.md
@@ -76,6 +76,22 @@ fmt.Println(val)
 // 3
 ```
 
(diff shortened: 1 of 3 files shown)


mattn:
Thank you

PR #38, opened 2016-08-15
improved error message for bad syntax

dhiltonp:
I found debugging bad syntax to be difficult, so I improved the error message.

Instead of `syntax error`, an error will be `syntax error occurred at or before l16:c5, symbol '{'`

The line calculation is a little weird - I couldn't figure out why l.pos.Line incremented by 2 on my systems and I'm concerned it may not be cross-platform.

It is working for me on both Linux and OSX.

### DIFF
--- a/parser/lexer.go
+++ b/parser/lexer.go
@@ -516,7 +516,9 @@ func (l *Lexer) Error(msg string) {
 func Parse(s *Scanner) ([]ast.Stmt, error) {
 	l := Lexer{s: s}
 	if yyParse(&l) != 0 {
-		return nil, l.e
+		badline := (l.pos.Line-1)/2
+		e := fmt.Errorf("%s at or before l%d:c%d, symbol '%s'", l.e, badline+1, l.pos.Column, l.lit)
+		return nil, e
 	}
 	return l.stmts, l.e
 }


mattn:
please explain what the bad code is wrong for debugging?

dhiltonp:
I'm not quite sure what you mean, but I have now tested the code with both CR/LF and LF line endings and the code works correctly, so I have removed the commented-out debugging statements.

mattn:
please show me your anko script.

dhiltonp:
I was porting some go code to anko for testing. I have fixed my code that wasn't working, and I didn't save the code with syntax errors - some were bad assignment (`:=`, not `=`), one was in a `for` conversion, one was due to my not converting a golang struct to an anko map.

For testing I have been inserting invalid mathematical expressions; `:=` instead of `=`, or adding a `*` on a line by itself to any script.

mattn:
I don't understand what you change and what you fix.

dhiltonp:
Ah! It's because I'm looking at it from an embedding perspective, not a CLI perspective. I need to move the change elsewhere to just affect the embedded error.

I had thought that was the highest level where the position data was available, but apparently this isn't the case - you can cast the error to a more specific error type.

Now that I understand anko.go better, I can see this isn't the ideal approach.

I have discovered a related bug and will open a ticket for that, as I couldn't find the root cause.

zhangyc310:
I'm crazy for the worst anko script , I use it as an emberd script in my soft.
 I run the emberd script ,when an err , it only says "Execute error: syntax error", 
where is the syntax error ? 
No one tell you !

mattn:
Please see this part https://github.com/mattn/anko/blob/fcefaff5544d88907bf86205aac7f06c58e3d412/anko.go#L149-L150

PR #77, opened 2017-10-30
Removed unnecessary toFloat64 & toInt64

MichaelS11:
If wanted, just removed some toFloat64 and toInt64 calls that should not be necessary.


mattn:
Thank you

PR #79, opened 2017-10-30
Added env.Copy() function to deep copy a scope

c4s4:
I need to copy the state of the VM, so I implemented this Copy() function

### DIFF
--- a/vm/env.go
+++ b/vm/env.go
@@ -260,13 +260,17 @@ func (e *Env) Execute(src string) (reflect.Value, error) {
 
 // Copy makes a deep copy of the scope
 func (e *Env) Copy() *Env {
-	b := false
 	envCopy := deepcopy.Copy(e.env).(map[string]reflect.Value)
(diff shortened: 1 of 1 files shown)


MichaelS11:
What about handling parents or children?

c4s4:
You're right, I have made a new pull request that recursively copies parents. For children, I have no idea how to do it or even if this is a good idea.

MichaelS11:
Should be able to do something very similar for the children.

TrueFurby:
To copy children as well you would need some global map of Envs to loop through and find each that has parent set to the one being copied, since Env only keeps info about it's parent but not vice-versa. 

Altough, I think copying parents should be enough. I someone wants to copy and Env they can still use it on the child directly.

MichaelS11:
Each parent could keep an slice of all it's children, would not have to be global.

If one is going to implement copy, shouldn't it be a full copy? Otherwise call it copyEnvAndParents or something like that.

MichaelS11:
@mattn 

The source repo was deleted, probably can just close this PR.

PR #80, opened 2017-10-31
Split vm code into three additional files

MichaelS11:
No code change on on split commit

Added CanAddr check for Addr
Major update to env_test, 99% coverage of env
Updated interrupt tests to remove sleep wait


mattn:
Cool!

PR #81, opened 2017-11-01
Added more tests. nil should now work better

MichaelS11:
Still have more comparison operator tests to do, but wanted to do a PR for what is done so far.


vm/env.go
```diff
@@ -160,15 +160,18 @@ func (e *Env) Get(k string) (reflect.Value, error) {
...
+	if ok {
+		var val reflect.Value
+		val, ok = v.(reflect.Value)
 		if !ok {
 			val = reflect.ValueOf(v)
 		}
+		e.Lock()
 		e.env[k] = val
+		e.Unlock()
 		return nil
```
mattn: I think this possible be race condition.

TrueFurby: I think situation can arise when two goroutines try to use Set and place between using RUnlock() and e.Lock() could be accessed by two subroutines, right?

mattn: Yes.

vm/env.go
```diff
@@ -223,9 +226,15 @@ func (e *Env) Define(k string, v interface{}) error {
 	if strings.Contains(k, ".") {
 		return fmt.Errorf("Unknown symbol '%s'", k)
 	}
-	val, ok := v.(reflect.Value)
-	if !ok {
-		val = reflect.ValueOf(v)
+	var val reflect.Value
+	if v == nil {
+		val = NilTypeNonPointer
```
TrueFurby: What is achieved by this? You can still set nil to variable now. What is the benefit of `NilTypeNonPointer`.

vm/env.go
MichaelS11: The read lock is a check if the variable is in the same env. If one did a define and set at the same time, one might set the variable in a different env than expected. This could happen even if there was a write lock on the whole function. If there are some Go routines running defines and some running sets on the same variable name in different parent/child env, there will always be a race condition on  which env or more than one env that variable name gets defined/set.

Example With pEnv & cEnv (even if set had global write lock):
GoR1: pEnv.Define A
GoR1: cEnv.Define A
GoR2: cEnv.Set A
Result: A set in cEnv 

GoR1: pEnv.Define A
GoR2: cEnv.Set A
GoR1: cEnv.Define A
Result: A set in pEnv

There should be no deadlock from this code change

vm/env.go
MichaelS11: This is the smallest change I could find that will return the correct nil value for a variable. If there is a better way, I would really like to know and understand.

TrueFurby: Could you provide example how it behaved before and with this change?

MichaelS11: Variables returned would be set to <invalid reflect.Value>instead of <nil>.

Here is one of the tests that would catch it:
{script: "a = b", input: map[string]interface{}{"b": nil}, runOutput: nil, ouput: map[string]interface{}{"a": nil, "b": nil}},

Here is the output from the test when the change is reverted:
outputName b - received: <invalid reflect.Value> expected: <nil> - script: a = b

vm/env.go
mattn: In this part, `ok` possibly become true at multiple groutine in same time?

https://github.com/mattn/anko/pull/81/files/574a908f79c43f50ce76f71ad3ede3dae2ff7cd4#diff-12238057aa85e299611b1bd13ffa555fR166

MichaelS11: Yes, if the variable is not Define yet, it would keep being false.

The difference between this updated Set function and the previous one is that in the previous version once Set gets its write lock any Define calls don't affect it. In this version, the Define call still affects the out come of the Set call in progress, depending on where the code is in the Set function. Both the previous version and this version have race conditions, depending on which function gets the locks first.  The winner of that race will affect if the variable will be set in this env scope or the parent scope, if there is a parent env.

So lets say two scripts are running in the same env with no parent scope, both are trying to set the same variable at the same time. Both get past the Set read lock and show it is false, and then they both return the Unknown symbol error. So both scripts call Define at the same time. Both hit the Define lock at the same time but only one of them will win. After one of the scripts sets the variable, the other one will override the variable. Both had to make the extra call to Define, but that should not be the usual case.

mattn:
```
#!anko

func foo() {
  var f = 1
  func() {
    #var f = 0
    f = 2
  }()
  print(f)
}

foo();
```
When this code is run, it show 2. And when un-comment the line, it show 1. Set will follow the parent scope to find f if exists in the scope. If the variable f is asynchronously set with Define on another goroutine, the Set will reference the parent 's env.

mattn:
This is problem of test script. It seems this PR failed.

https://travis-ci.org/mattn/anko/jobs/296008379

vm/env.go
MichaelS11: Are you you saying something is not working correctly with that code? If so, could you please point it out to me?

MichaelS11:
Looking into the issue.

vm/env.go
MichaelS11: Removed NilTypeNonPointer. Changed NilValue to be an inteface{} (nil) which should fix the issue and not cause this weirdness.

MichaelS11:
Fix issue with tests.
Removed the NilTypeNonPointer and change NillValue to no longer be a pointer so that values return can be interface{} (nil)
My apologizes for so many changes in one PR.

vm/env.go
MichaelS11: Added TestRaceDefineAndSetSameVariable, which shows the define/set race.

MichaelS11:
If this pull request is too messy, please just let me know, I will have time to redo it in a couple of weeks.

mattn:
Thank you. I'll look this later.

mattn:
Works good. Thanks.

PR #82, opened 2017-11-08
added an interrupt check statement on for loops so that we can break …

traetox:
If a script consists of "for{}" its not possible to break out of that statement because the loop statement has no sub statements.  I abstracted out the interrupt check to its own call to make it a little cleaner and added the check on the interrupt into each iteration of a for loop.

Its an edge case, but if you are running user supplied scripts, its possible for the user to livelock the VM in a way that you can't interrupt out of.

### DIFF
--- a/vm/env.go
+++ b/vm/env.go
@@ -65,6 +65,18 @@ func (e *Env) NewPackage(n string) *Env {
 	}
 }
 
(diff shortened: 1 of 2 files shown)


mattn:
Thank you

PR #83, opened 2017-11-09
Adding another interrupt statement to catch bad forloops

traetox:
Sorry, missed one.

the code
```
for i = 0; i < 10; i += 0 {} 
```
is also uninterruptible.  Added another interrupt catch to fix it


mattn:
Thank you

PR #84, opened 2017-11-09
Add external variable resolver to env.

floren:
This is anything which has the Get and Type functions, allowing the environment to use it to look up and fetch variables.

By adding this, I was able to use anko as a query language which can pull variables on-the-fly from external resources.

### DIFF
--- a/vm/env.go
+++ b/vm/env.go
@@ -9,6 +9,11 @@ import (
 	"github.com/mattn/anko/parser"
 )
 
+type EnvResolver interface {
+	Get(string) (reflect.Value, error)
+	Type(string) (reflect.Type, error)
+}
+
 // Env provides interface to run VM. This mean function scope and blocked-scope.
 // If stack goes to blocked-scope, it will make new Env.
 type Env struct {
@@ -17,6 +22,7 @@ type Env struct {
 	typ       map[string]reflect.Type
 	parent    *Env
 	interrupt *bool
+	external  EnvResolver
 	sync.RWMutex
 }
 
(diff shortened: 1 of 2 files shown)


mattn:
Could you plaease add?

2017/11/10 午前4:38 "John Floren" <notifications@github.com>:

This is anything which has the Get and Type functions, allowing the
environment to use it to look up and fetch variables.

By adding this, I was able to use anko as a query language which can pull
variables on-the-fly from external resources.
------------------------------
You can view, comment on, or merge this pull request online at:

  https://github.com/mattn/anko/pull/84
Commit Summary

   - Add external to env. This is anything which has the Get and Type
   functions, allowing

File Changes

   - *M* vm/env.go <https://github.com/mattn/anko/pull/84/files#diff-0>
   (30)

Patch Links:

   - https://github.com/mattn/anko/pull/84.patch
   - https://github.com/mattn/anko/pull/84.diff

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/mattn/anko/pull/84>, or mute the thread

floren:
Hi mattn, what would you like me to add?

mattn:
Ah. sorry. Could you please add test?

floren:
I've added a test, I hope it demonstrates the intent a bit and shows how it works. The idea is that this will make it easier to let anko access variables from other sources, you could even make an interface to let it read things out of a key/value store or database!

mattn:
> The idea is that this will make it easier to let anko access variables from other sources

Looks nice.

mattn:
Thank you

MichaelS11:
Note that the tests do not cover the following code inside the if:

        if e.external != nil {
                v, err := e.external.Get(k)
                if err == nil {
                        if v.CanAddr() {
                                return v.Addr(), nil
                        }
                }
        }

PR #85, opened 2017-11-14
Implicit conversions in equality comparisons

floren:
This is already sort-of implemented for >, <, >=, <=, and the arithmetic operations, but this adds the capability to evaluate things like:

"1.0" == 1 (evaluates to true)
1.1 == "1.0" (evaluates to false)
"7" == 7.2 (evaluates to true, "7" is an integer so we cast int(7.2) -> 7)

### DIFF
--- a/vm/vm.go
+++ b/vm/vm.go
@@ -120,11 +120,34 @@ func equal(lhsV, rhsV reflect.Value) bool {
 	if !lhsV.IsValid() || !rhsV.IsValid() {
 		return true
 	}
+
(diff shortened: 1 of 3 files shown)


mattn:
Thank you

MichaelS11:
Not sure if converting strings to ints/floats is necessarily a good idea for implicit conversions. Wouldn't it be better to have a to number function for strings?

https://www.w3schools.com/jsref/jsref_number.asp

PR #90, opened 2017-11-17
Function to walk the AST

traetox:
I added a function that will walk a statement list and call back to provided functions for each statement or expression.

This is useful for doing some basic syntax validation prior to actually running the AST.  We use this to sweep the AST and ensure that any calls to our golang implemented functions are providing the appropriate number of arguments at a parse phase rather than executing and failing at execution.

An example usage:

if err := ast.WalkStmts(stmts, nil, func(e ast.Expr) error{
switch e := e.(type) {
case *ast.CallExpr:
   if (e.SubExpr) != FunctionToLookupParamemterCount(e.Name) {
      return errors.New("Invalid parameter count on "+e.Name)
   }
}
}); err != nil {
   return err
}

### DIFF
--- a/ast/util/walk.go
+++ b/ast/util/walk.go
@@ -1,17 +1,19 @@
-package ast
+package astutils
 
 import (
 	"fmt"
 	"reflect"
+
+	"github.com/mattn/anko/ast"
 )
 
-type WalkStatementFunc func(Stmt) error
-type WalkExpressionFunc func(Expr) error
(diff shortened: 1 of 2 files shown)


mattn:
Thanks your contribution. This function is embed into the binary. Could you move this into `ast/util` ?

traetox:
Absolutely. That's a good pickup and I appreciate the call out.

I'll move it and reissue the PR.

PR #92, opened 2017-11-27
Updated len, return nil correction, added tests

MichaelS11:


MichaelS11:
Added more tests

mattn:
Thank you

PR #97, opened 2017-12-06
 Updated LetsExpr and LetsStmt, added tests

MichaelS11:
Please note this does change how Run output value behavior for let statements when they have multiple values, it will only return the last value set.

### DIFF
--- a/vm/vmExpr.go
+++ b/vm/vmExpr.go
@@ -541,29 +541,20 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
 		}
 		return invokeLetExpr(e.Lhs, rv, env)
 	case *ast.LetsExpr:
-		rv := NilValue
 		var err error
-		vs := []interface{}{}
-		for _, rhs := range e.Rhss {
-			rv, err = invokeExpr(rhs, env)
+		rvs := make([]reflect.Value, len(e.Rhss))
+		for i, rhs := range e.Rhss {
+			rvs[i], err = invokeExpr(rhs, env)
 			if err != nil {
 				return NilValue, NewError(rhs, err)
 			}
-			if rv == NilValue {
-				vs = append(vs, nil)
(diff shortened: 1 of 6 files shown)


vm/vmExpr.go
```diff
@@ -541,37 +541,28 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+			if v.IsValid() && v.Kind() == reflect.Interface && !v.IsNil() {
 				v = v.Elem()
 			}
 			_, err = invokeLetExpr(lhs, v, env)
 			if err != nil {
 				return NilValue, NewError(lhs, err)
 			}
 		}
-		return rvs, nil
+		return rvs[len(rvs)-1], nil
```
mattn: what this mean?

vm/vmStmt.go
```diff
@@ -43,68 +43,56 @@ func RunSingleStmt(stmt ast.Stmt, env *Env) (reflect.Value, error) {
...
-			if i < len(rvs) {
-				env.Define(name, rvs[i])
-				result = append(result, rvs[i].Interface())
+			if i >= len(rvs) {
+				break
 			}
+			env.Define(name, rvs[i])
 		}
-		return reflect.ValueOf(result), nil
+		return rvs[len(rvs)-1], nil
```
mattn: ditto

MichaelS11:
This PR changes how multiple variable set expressions return run output. For example:
```
a, b = 1, 2
```
Run output before PR:
```
[] interface{}{int64(1), int64(2)}
```
Run output after PR:
```
int64(1)
```

This goes for multiple variable var statements as well.

MichaelS11:
Updated interrupt, changed to be a flag that can be set and unset for environment group.

mattn:
Ah, I understand what this mean.

mattn:
Thank you

PR #99, opened 2017-12-09
Updated ItemExpr, added, tryToInt, updated isNil, added tests

MichaelS11:
vm coverage now at 57.3% of statements


mattn:
SGTM, you change it to make array index out of range instead of returning nil. right?

MichaelS11:
Yes

mattn:
Thank you

PR #100, opened 2017-12-12
 Updated ast.MemberExpr, equal, and added more tests

MichaelS11:


MichaelS11:
@mattn ,

Any thoughts on this?

mattn:
Please note what you want to change. AFAICS, you seems to want to change `nil == nil` and add `kindOf` ?

MichaelS11:
Added toInt, updated error codes, changed so return can be used in main, added kindOf, updated member expressions and index expressions. Those are the bigger changes. The update to IsValid seemed fairly minor, not sure if it really changed much.

builtins/regexp/regexp.go
```diff
@@ -8,7 +8,7 @@ import (
 )
 
 func Import(env *vm.Env) *vm.Env {
-	m := env.NewPackage("sort")
+	m := env.NewPackage("regexp")
```
mattn: good catch. thank you

MichaelS11: Welcome :)

MichaelS11:
Note that the changes are grouped into there own commit. The addition of new tests or test changes are also in there own commit. The hope was that would make it easier to review. Please let me know if there is anything else I can do to help make the changes easier to review.

mattn:
LGTM. Thank you. In next time for PR, please note what you want to change in description.

PR #103, opened 2017-12-22
Change output from reflect.Value to interface

MichaelS11:
I think that interface is a more common and easier to use way to get output out. Most people are not as familiar with reflect.Value and it can be more difficult to use. People can often use the interface output as is, or if not, they can do a type assertion to the type they expect/need.

I also think this will abstract the internals of Anko so if it is wanted to change how variables are stored inside the env, the outside function returns will not have to change again.

Please note this breaks backwards compatibility for:
Get, Set, Define, DefineGlobal, Run, and RunSingleStmt


MichaelS11:
Happy Holidays! :)

mattn:
Thanks

PR #107, opened 2018-01-02
Fix map key not present, added tests to Travis, added more tests

MichaelS11:


MichaelS11:
Improved automatic append

MichaelS11:
Fix panic on uninitialized map

vm/vm.go
```diff
@@ -190,3 +191,35 @@ func equal(lhsV, rhsV reflect.Value) bool {
...
 	return reflect.DeepEqual(lhsV, rhsV)
 }
+
+func getMapIndex(key reflect.Value, aMap reflect.Value) reflect.Value {
+	if !aMap.IsValid() || aMap.IsNil() {
+		return NilValue
+	}
+
+	keyType := key.Type()
+	if keyType.String() == "interface {}" && aMap.Type().Key().String() != "interface {}" {
```
TrueFurby: String `interface {}` seems to be used quite often, might it be better to define it as constant?

MichaelS11: Performance wise, doesn't the compiler basically do that already?

Or where you thinking human reading wise it would be better?

TrueFurby: You are right the compiler probably already does that. However using a constant not only makes reading the code better, but also guards developer from making accidental mistake in one of the strings. 

For performance wise reasons the best would be to use constants of specific type and compare the types instead of comparing strings. For example:

```go
var emptyInterface = reflect.TypeOf(interface{})
if key.Type == emptyInterface { ... }
```

MichaelS11: Great idea removing string compare! :)

New commit pushed with change.

mattn:
Thank you

PR #115, opened 2018-01-23
Index of string now returns string instead of rune

MichaelS11:


mattn:
You will add more test to check type of `str[0]`?

MichaelS11:
Added a couple more tests. Was that what you wanted? If it does not, not clear what is wanted.

MichaelS11:
Added interface type
Added TestJson
Added tests to TestKindOf

MichaelS11:
@mattn ,
I think the tests provided should work to make sure a string is returned. Are you alright with them?

MichaelS11:
@mattn ,
Is this good?

builtins/core.go
```diff
@@ -300,10 +300,11 @@ func Import(env *vm.Env) *vm.Env {
...
 	})
 
-	env.DefineType("int64", int64(0))
-	env.DefineType("float64", float64(0.0))
+	env.DefineType("interface", reflect.ValueOf([]interface{}{int64(1)}).Index(0).Type())
 	env.DefineType("bool", true)
-	env.DefineType("string", "")
+	env.DefineType("int64", int64(1))
+	env.DefineType("float64", float64(1))
+	env.DefineType("string", "a")
```
mattn: you will add rune?

MichaelS11:
Added rune

mattn:
Thanks so much

PR #119, opened 2018-02-07
Added for key, value in map

MichaelS11:


MichaelS11:
Fix ++/-- on arrays
Updated ast.AssocExpr
Bool true is now equal to 1/1.0 instead of 0

MichaelS11:
env is now at 100% coverage again!

Fixed Addr for external
Updated and added test to env external resolver
Added TestGetInvalid
Added tests to TestFunctions

mattn:
Thank you

PR #123, opened 2018-02-23
**Breaking change for types** plus more

MichaelS11:
**Breaking change for types**

Make arrays last two args now optional
Changed type to be an expression
Type expression converts to string for type lookup
DefineType is now local scope
Added DefineReflectType
Added DefineGlobalType & DefineGlobalReflectType
Added make multidimensional array support
Improved multidimensional array support
Added AddPackage


MichaelS11:
Packages moving out of core coming soon now that types are local scope.

parser/parser.go.y
```diff
@@ -249,6 +250,16 @@ stmt_default :
 		$$ = &ast.DefaultStmt{Stmts: $4}
 	}
 
+array_count :
+	array_count '[' ']'
+	{
+		$$ = ast.ArrayCount{Count: $1.Count + 1}
```
MichaelS11: Is there a way to do $1.Count++ ? I was having issues getting that to work.

mattn:
Hmm, it is impossible to change string literals in new/make?

MichaelS11:
My first reaction was also to keep literals for new/make, but then I started thinking about it more. Having expressions for new/make can be really useful and powerful with very little overhead. A simple example:

```
...
a = make([]typeOf(b))
...
```

If you did that with literals it would have to be a case statement, maybe a really large case statement depending on how many types.

MichaelS11:
Looking into creating a more general SortArray for core

MichaelS11:
Working on refactor for function code to see about handling sort functions

MichaelS11:
This should be ready. Now can do sort functions inside of VM.

mattn:
Well, is not possible to implement non-quoted like `make([]bool)` ?

MichaelS11:
Yes it is possible, but probably not good to have the string literals syntax as well the expression syntax . Probably should choice one or the other. I personally like the expression syntax.

MichaelS11:
Is sounds like you want me to change it back to string literals syntax. Is that what you would like me to do?

mattn:
>Is sounds like you want me to change it back to string literals syntax. Is that what you would like me to do?

Yes, I want to keep compatibility for Go as possible.

MichaelS11:
Changed to string literal

MichaelS11:
What do you think?

MichaelS11:
This good to go?

mattn:
Well, as I mentioned in above, I hope to keep implementation of `make([]string 3)`.

MichaelS11:
One can do any of the following:
```
make([]string)
make([]string, 3)
make([]string, 1, 3)
make([][]string)
make([][]string, 3)
make([][]string, 1, 3)
```
etc...

It can not do `make([]string 3)` without comma.

mattn:
comma is my mistake. but it seems that some string literals are remaining in example?

mattn:
Big Thanks.

PR #132, opened 2018-03-21
Updated DeleteExpr to handle any map key type

MichaelS11:
Also many fixes for Go Report Card


MichaelS11:
@mattn 
Side note, any thoughts about breaking change of renaming errors to have Err in front? For example BreakError to ErrBreakError.

mattn:
Many Thanks.

PR #140, opened 2018-03-22
 Improved convertReflectValueToType

MichaelS11:
Updated Interrupt tests
Added Interface check to DeleteExpr

### DIFF
--- a/vm/vmExprFunction.go
+++ b/vm/vmExprFunction.go
@@ -148,7 +148,7 @@ func callExpr(callExpr *ast.CallExpr, env *Env) (rv reflect.Value, err error) {
 		for i, expr := range callExpr.SubExprs {
 			if addrExpr, ok := expr.(*ast.AddrExpr); ok {
 				if identExpr, ok := addrExpr.Expr.(*ast.IdentExpr); ok {
-					invokeLetExpr(identExpr, args[i].Elem().Elem(), env)
+					invokeLetExpr(identExpr, args[i].Elem(), env)
 				}
 			}
 		}
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -326,6 +326,15 @@ func TestLen(t *testing.T) {
 	runTests(t, tests)
 }
 
+func TestReferencingAndDereference(t *testing.T) {
+	os.Setenv("ANKO_DEBUG", "1")
+	tests := []testStruct{
+		// TOFIX:
+		// {script: `a = 1; b = &a; *b = 2; *b`, runOutput: int64(2), output: map[string]interface{}{"a": int64(2)}},
+	}
+	runTests(t, tests)
+}
+
 func runTests(t *testing.T, tests []testStruct) {
 	var value interface{}
 loop:
--- a/vm/vm_test.go
+++ b/vm/vm_test.go
@@ -456,7 +456,7 @@ for {
 `,
 		`
 a = []
-for i = 0; i < 10000; i++ {
+for i = 0; i < 100000; i++ {
 	a += 1
 }
 closeWaitChan()
@@ -465,7 +465,7 @@ for i in a {
 `,
 		`
 a = []
-for i = 0; i < 10000; i++ {
+for i = 0; i < 100000; i++ {
 	a += 1
 }
 closeWaitChan()
@@ -488,7 +488,7 @@ for i = 0; true; nil {
 `,
 		`
 a = {}
-for i = 0; i < 10000; i++ {
+for i = 0; i < 100000; i++ {
 	a[toString(i)] = 1
 }
 closeWaitChan()
@@ -497,7 +497,7 @@ for i in a {
 `,
 		`
 a = {}
-for i = 0; i < 10000; i++ {
+for i = 0; i < 100000; i++ {
 	a[toString(i)] = 1
 }
 closeWaitChan()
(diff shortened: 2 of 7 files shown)


vm/vmExprFunction.go
```diff
@@ -148,7 +148,7 @@ func callExpr(callExpr *ast.CallExpr, env *Env) (rv reflect.Value, err error) {
 		for i, expr := range callExpr.SubExprs {
 			if addrExpr, ok := expr.(*ast.AddrExpr); ok {
 				if identExpr, ok := addrExpr.Expr.(*ast.IdentExpr); ok {
-					invokeLetExpr(identExpr, args[i].Elem().Elem(), env)
+					invokeLetExpr(identExpr, args[i].Elem(), env)
```
MichaelS11: @mattn 
So this reverts the previous commit you did. May I ask what you were trying to fix? Can I add a test for what you were trying to fix?

I added a test that will cause a panic if this is not change.

vm/vm_test.go
```diff
@@ -456,7 +465,7 @@ for {
 `,
 		`
 a = []
-for i = 0; i < 10000; i++ {
+for i = 0; i < 100000; i++ {
```
mattn: please separate changes for TestInterrupts

vm/vmExprFunction.go
mattn: first Elem() is for dereference the pointer-variable. second one is for dereference the value that was set via v.Set (should be reflect.Interface).

vm/vm_test.go
MichaelS11: Reverted TestInterrupts count change

vm/vm_test.go
```diff
@@ -536,10 +545,8 @@ func runInterruptTest(t *testing.T, script string) {
 	}()
 
 	_, err = env.Execute(script)
-	if err == nil {
-		t.Errorf("Execute error - received %v - expected: %v", err, InterruptError)
-	} else if err.Error() != InterruptError.Error() {
-		t.Errorf("Execute error - received %v - expected: %v", err, InterruptError)
+	if err == nil || err.Error() != InterruptError.Error() {
```
mattn: please fix conflict.

vm/vmExprFunction.go
MichaelS11: But it might not be a interface, right?

MichaelS11:
Merge fixed

vm/vmExprFunction.go
mattn: sure.

vm/vm_test.go
```diff
@@ -581,9 +584,7 @@ func TestInterruptConcurrency(t *testing.T) {
 	for i := 0; i < 100; i++ {
 		go func() {
 			_, err := env.Execute("for { }")
-			if err == nil {
-				t.Errorf("Execute error - received %v - expected: %v", err, ErrInterrupt)
-			} else if err.Error() != ErrInterrupt.Error() {
+			if err != ErrInterrupt {
```
mattn: ditto

vm/vm_test.go
```diff
@@ -564,9 +569,7 @@ func TestInterruptConcurrency(t *testing.T) {
 	waitGroup.Wait()
 
 	_, err := env.Execute("for { }")
-	if err == nil {
-		t.Errorf("Execute error - received %v - expected: %v", err, ErrInterrupt)
-	} else if err.Error() != ErrInterrupt.Error() {
+	if err != ErrInterrupt {
```
mattn: ditto

vm/vm_test.go
```diff
@@ -551,9 +558,7 @@ func TestInterruptConcurrency(t *testing.T) {
 	for i := 0; i < 100; i++ {
 		go func() {
 			_, err := env.Execute("for { }")
-			if err == nil {
-				t.Errorf("Execute error - received %v - expected: %v", err, ErrInterrupt)
-			} else if err.Error() != ErrInterrupt.Error() {
+			if err != ErrInterrupt {
```
mattn: ditto

vm/vm_test.go
```diff
@@ -536,10 +545,8 @@ func runInterruptTest(t *testing.T, script string) {
 	}()
 
 	_, err = env.Execute(script)
-	if err == nil {
-		t.Errorf("Execute error - received %v - expected: %v", err, ErrInterrupt)
-	} else if err.Error() != ErrInterrupt.Error() {
-		t.Errorf("Execute error - received %v - expected: %v", err, ErrInterrupt)
+	if err == nil || err.Error() != ErrInterrupt.Error() {
```
mattn: Let's remove this part at this PR.

MichaelS11: May I ask why? All that is doing it removing an extra if statement.

mattn: It makes it difficult for me to find issue that problem related to a bug.

mattn:
merged https://github.com/mattn/anko/pull/142

vm/vm_test.go
MichaelS11: @mattn 

Would be happy to understand why this change makes it difficult to find the issue. If you could please let me know why, that would be great.

Here is what I could see:

The job https://travis-ci.org/mattn/anko/jobs/356702744 said:
--- FAIL: TestInterrupts (0.11s)
	vm_test.go:540: Execute error - received <nil> - expected: Execution interrupted

Can look at line 540 here:
https://github.com/mattn/anko/blob/72fa4855a49189e82452215a4cc4cb22b16b2b8c/vm/vm_test.go#L540

That line is in the function runInterruptTest, which will runs these tests:
https://github.com/mattn/anko/blob/72fa4855a49189e82452215a4cc4cb22b16b2b8c/vm/vm_test.go#L445-L507

So the change in the if statements will not make it any easier to figure out which test was the problem, right?

MichaelS11:
@mattn 

Could you please let me know why some of this PR was dropped? Or if there is a general change on how you would like PRs to be done, opened up a issue about it.

mattn:
Please do not feel wrong. I just want to separate PRs since I prefer that one PR is fix one issue. This is useful to find out what was wrong. I developed anko as I feel as. But now anko seems to be depended on some products. So I decided this. Please don't feel wrong meaning.

MichaelS11:
No problem, happy to make PRs smaller when possible. Maybe next time it would be possible to share about the change ahead of time to make it easier to adapt?

Chyroc:
Everyone is great

PR #144, opened 2018-03-22
add v, ok = m[k]

### LINKED ISSUE #105: Getting nil from map
Following script :

```
fmt=import("fmt")

foo = {"foo": "bar", "spam": "eggs"}
if foo["toto"] == nil {
    fmt.Println("nil")
} else {
    fmt.Println("non nil")
}
fmt.Printf("foo['toto']: %v\n", foo["toto"])
```

Produces output:

```
$ anko test.ank 
non nil
foo['toto']: <nil>
```

Which seems broken!

mattn:
closes #105


mattn:
This is wrong approach.

PR #145, opened 2018-03-22
add statement for v, ok = m[k]

### LINKED ISSUE #105: Getting nil from map

mattn:
closes #105

### DIFF
--- a/ast/stmt.go
+++ b/ast/stmt.go
@@ -125,3 +125,11 @@ type LetsStmt struct {
 	Operator string
 	Rhss     []Expr
 }
+
+// LetMapItemStmt provide statement of let for map item.
+type LetMapItemStmt struct {
+	StmtImpl
+	Lhss     []Expr
+	Operator string
+	Rhss     []Expr
+}
--- a/ast/stmt.go
+++ b/ast/stmt.go
@@ -129,7 +129,6 @@ type LetsStmt struct {
 // LetMapItemStmt provide statement of let for map item.
 type LetMapItemStmt struct {
 	StmtImpl
-	Lhss     []Expr
-	Operator string
-	Rhs      Expr
+	Lhss []Expr
+	Rhs  Expr
 }
--- a/ast/stmt.go
+++ b/ast/stmt.go
@@ -131,5 +131,5 @@ type LetMapItemStmt struct {
 	StmtImpl
 	Lhss     []Expr
 	Operator string
-	Rhss     []Expr
+	Rhs      Expr
 }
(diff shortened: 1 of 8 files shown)


MichaelS11:
Would it be possible to add the new key tests in a new test function maybe called TestExistenceOfKeyInMaps and put it in vmArraysAndMaps_test.go instead?

vm/vmExpr.go
```diff
@@ -256,6 +256,9 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
 			return v.FieldByIndex(field.Index), nil
 		case reflect.Map:
 			v = getMapIndex(reflect.ValueOf(e.Name), v)
+			if v == zeroValue {
```
MichaelS11: Add comment back in?
// Note if the map is of reflect.Value, it will incorrectly return nil when zero value

ast/stmt.go
```diff
@@ -125,3 +125,11 @@ type LetsStmt struct {
 	Operator string
 	Rhss     []Expr
 }
+
+// LetMapItemStmt provide statement of let for map item.
+type LetMapItemStmt struct {
+	StmtImpl
+	Lhss     []Expr
+	Operator string
+	Rhss     []Expr
```
MichaelS11: Does Rhss need to be a slice?

vm/vmStmt.go
```diff
@@ -117,6 +117,28 @@ func runSingleStmt(stmt ast.Stmt, env *Env) (reflect.Value, error) {
...
+		}
+		var rvs []reflect.Value
+		if rv != zeroValue {
+			rvs = []reflect.Value{rv, trueValue}
+		} else {
+			rvs = []reflect.Value{nilValue, falseValue}
+		}
+		for i, lhs := range stmt.Lhss {
+			v := rvs[i]
+			if v.IsValid() && v.Kind() == reflect.Interface && !v.IsNil() {
```
MichaelS11: FYI, Kind will return Invalid if v is not valid, could remove the v.IsValid() && part.
https://golang.org/pkg/reflect/#Value.Kind

MichaelS11:
Looks good.

Was curious, is Operator needed in LetMapItemStmt? Was there a future case where it would not be set to =?

If you want, I will add some more tests after it is merged.

PR #165, opened 2018-03-23
add appendNumberAndPoint

Chyroc:


MichaelS11:
I now see what this is doing, very cool. The concept is great. :)

I am not sure if it will work in all cases. In particular, packages that have new methods adding to them, like sort as an example. There is a sort 1.8 build. Also note that some packages have appengine build constraints/tags.

It should at least be able to use as a good starting template for checking what we have or adding new packages.

mattn:
Thank you

Issue #49, opened 2017-01-31
math/big error on arm


njaremko:
When attempting to build on arm, I get this:
`src/github.com/mattn/anko/builtins/math/big/big.go:19: constant 4294967295 overflows int`

and the build fails.

mattn:
Could you please show me your code reproduce?

njaremko:
No code, just running `go get github.com/mattn/anko`

mattn:
Hmm, it seems a bug of go. could you please file issue on https://github.com/golang/go/issues/new ?

mattn:
But I can do workaround that removing the part of line until go fix the issue. Could you please try to figure out what line should be removed.

njaremko:
Line 19 of big.go

mattn:
Yes, it's only 1 line?

njaremko:
Yep

mattn:
removed math.MaxPrec temporary (until fixing this go/arm issue), and I keep to open this issue until it.

mattn:
@njaremko Could you please file your issue? https://github.com/golang/go/issues/new

MichaelS11:
This may not have been an issue with Go, nor with arm. Maybe this was just issue with a 32 bit system.

From Big:
MaxPrec = math.MaxUint32 // largest (theoretically) supported precision; likely memory-limited

So a unit32 const was being put into a int which was an int32, which means the value was too big.

Maybe now everything will be fine now that we are using the new packages? I do not have a 32 bit system to test. Anyone have a 32 bit system to test?

MichaelS11:
@mattn any thoughts?

PR #167, opened 2018-03-26
add generate-package-code

Chyroc:


Chyroc:
At present, we only take `bytes` as an example of pr. If you think ok, I will add other packages.

MichaelS11:
If I understand correctly, you want to add an example how to use package?

If so, I think it should be done similar to this:

https://github.com/mattn/anko/blob/master/vm/exampleBasicOperators_test.go

I will do one for sort soon.

Chyroc:
No, I mean maybe these standard packages can be done by generating code, because each package may have updates, and it is considered difficult to detect and update the code.
Then I just give the generated code and use `bytes` as the sample code. If I can, I will use the code to generate the code under the wrong package.

MichaelS11:
I now see what this is doing, very cool. The concept is great. :)

I am not sure if it will work in all cases. In particular, packages that have new methods adding to them, like sort as an example. There is a sort 1.8 build. Also note that some packages have appengine build constraints/tags.

It should at least be able to use as a good starting template for checking what we have or adding new packages.

mattn:
It is difficult to create a portable package file using this tool. Because syscall has different methods depending on OS. So I think that it must have a dictionary or whitelist.

Chyroc:
can use makefile to make only specific packages use the generated code

MichaelS11:
Might be better for how it is now to remove the makefile and just run it manually as wanted. Great tool, can do lots with it, just not full automatic.

MichaelS11:
@Chyroc 
Could you please fix bytes?

Chyroc:
I have resolved the conflict, but there are still some problems.

Columns such as 
```
	json.Number // type Number string
	time.Nanosecond // const Nanosecond  Duration = 1
```
 are a type or a var, but they are not structs, so they cannot use `AAA{}`. At the moment I have added them to the blacklist, but I don't think this is a good way. Maybe we should support this type, but what should we do?

MichaelS11:
time.Nanosecond is already added:
https://github.com/mattn/anko/blob/master/packages/time.go

MichaelS11:
I think something like this will work:
	PackageTypes["encoding/json"] = map[string]interface{}{
		"Number": json.Number(""),
	}

Also can check out Signal:
https://github.com/mattn/anko/blob/master/packages/os.go

MichaelS11:
Note that the build tests are failing because json.Valid was not added until go 1.9

Chyroc:
I noticed, I will fix him, I'm sorry to be busy recently

Chyroc:
I have a problem: The type of import is now through:
1, `bytes.Buffer{}` imported in this way,
2, the type of non-structure is imported in this way
```
var signal os.Signal
reflect.TypeOf(&signal).Elem()
```

Whether to continue this way, or unified as the second way

mattn:
interface can be nil. For example, This is typed nil.

```
v := io.Writer(nil)
```

MichaelS11:
A note, I think this code should try to handle all the cases. This code is great for generating a template that can be updated by hand.

Chyroc:
how to deal with `math.MaxUint64`, it return `constant 18446744073709551615 overflows int` in my machine

MichaelS11:
Should be able to just convert it to a uint64.

https://github.com/golang/go/issues/19621
https://github.com/golang/go/issues/23086

MichaelS11:
Have you read up on Build Constraints?

https://golang.org/pkg/go/build/

Looks like there are extra Build Constraints that are not wanted.

For example:

// +build go1.8,!go1.10

Which means build if (1.8 or greater) OR (not 1.10 or great), which probably not what we are going for.

MichaelS11:
Also there is the custom build tag appengine which I believe is used to create the engine for the web site. So `// +build !appengine` needs to be accounted for.

packages/errors.go
```diff
@@ -8,4 +8,6 @@ func init() {
 	Packages["errors"] = map[string]interface{}{
 		"New": errors.New,
 	}
+	PackageTypes["errors"] = map[string]interface{}{}
```
MichaelS11: Empty PackageTypes can be removed.

packages/math.110.go
```diff
@@ -0,0 +1,19 @@
...
+package packages
+
+import (
+	"math"
+)
+
+func init_math() {
+
+	Packages["math"]["Erfcinv"] = math.Erfcinv
+
```
MichaelS11: Not a big deal, but would be nice to remove the extra spaces.

packages/os.go
```diff
@@ -16,10 +26,12 @@ func init() {
 		"Create":            os.Create,
 		"DevNull":           os.DevNull,
 		"Environ":           os.Environ,
+		"ErrClosed":         os.ErrClosed,
 		"ErrExist":          os.ErrExist,
 		"ErrInvalid":        os.ErrInvalid,
 		"ErrNotExist":       os.ErrNotExist,
 		"ErrPermission":     os.ErrPermission,
+		"Executable":        os.Executable,
```
MichaelS11: Some of the added function should be added to `// +build !appengine` like Executable, Unsetenv, and etc.

packages/sort.go
```diff
@@ -16,26 +17,35 @@ func (s SortFuncsStruct) Less(i, j int) bool { return s.LessFunc(i, j) }
...
 		"Float64sAreSorted": sort.Float64sAreSorted,
 		"Ints":              sort.Ints,
 		"IntsAreSorted":     sort.IntsAreSorted,
 		"IsSorted":          sort.IsSorted,
+		"Reverse":           sort.Reverse,
 		"Search":            sort.Search,
 		"SearchFloat64s":    sort.SearchFloat64s,
 		"SearchInts":        sort.SearchInts,
 		"SearchStrings":     sort.SearchStrings,
+		"Slice":             sort.Slice,
```
MichaelS11: Slice* should be added to `// +build go1.8`

Chyroc:
Yes, I read ~
The comma means `and` and the space means `or`, so the `// +build go1.8,!go1.10` meaning here is 1.8 and above, but does not include 1.10

For example, the function `init_math`:
`// +build go1.8,!go1.10` will build under 1.8/1.9 in file `packages/math.18.go`
`// +build go1.10` will build under 1.10 in file `packages/math.110.go`
and they called by file `packages/math.go`

Take the function `init_encoding_json` for example:
`// +build go1.8,!go1.9,!go1.10` will build under 1.8 in file `packages/encoding.json.18.go`
`// +build go1.9 go1.10` will build under 1.9/1.10 in file `packages/encoding.json.19.110.go`
and they called by file `packages/encoding.json.go`

MichaelS11:
Also, how are you checking what version each of the newly adding functions and types are in?

MichaelS11:
Looking at json.Valid, that should be // +build go1.9 and the other file should be // +build !go1.9.

Chyroc:
Maybe we can run `./cmd/generate-std-packages/run.sh` in the ci test. If there is a change in the code, there is a new file, and ci should fail.
Of course, it can also be performed manually on a regular basis (so as not to block other people's pr)

MichaelS11:
While we are only testing 1.8, 1.9, 1.10 in travis, still want to try to be compatible to lower versions when/if possible. That means trying to figure out which version functions and types were added, for all the newly added functions and types.

MichaelS11:
Note https://github.com/mattn/anko/issues/49 for overflows int issue.

packages/errors.go
Chyroc: done.

packages/os.go
Chyroc: Ok. Then just ask, how to know the list of `// +build !appengine`

Chyroc:
Should we first create maps of all packages and then add functions or variables one by one.
 then add
```
l.Lock()
if Packages["name"] = nil {
Packages["name"]  = make(map[string]interface{})
}
l.Unlock()
```
This avoids the need to create two file builds similar to the appengine environment

MichaelS11:
We are adding functions or variables one at a time, don't think a lock is needed. Not seeing how this avoids the need for build files. 

If you want to have only have two appengine build files, can have those file sets a flag, then use that flag to set the data.

Chyroc:
run such code to generate code:
```bash
#!/bin/bash

mkdir go && cd go

for i in 1.6.4 1.7.6 1.8.7 1.9.6 1.10.3
do
   echo "download go $i ing."
   wget https://dl.google.com/go/go$i.linux-amd64.tar.gz && tar zxvf go$i.linux-amd64.tar.gz && mv go $i
   rm go$i.linux-amd64.tar.gz
   echo "done."
done

cd ..

find packages | grep -vE '_test|Engine' | grep '/' | xargs rm

go run cmd/generate-std-packages/*.go -go16 ./go/1.6.4/bin/go -go17 ./go/1.7.6/bin/go -go18 ./go/1.8.7/bin/go -go19 ./go/1.9.6/bin/go -go110 ./1.10.3/bin/go && gofmt -s -w ./packages/*

```

Chyroc:
@MichaelS11 @mattn please see again, thank you.

MichaelS11:
I leave this to @mattn 

I was fine with the simple starting template and making changes by hand.

mattn:
I don't have strong opnion, Is it possible to make composite literal like?

```
Packages["bytes"] = map[string]interface{}{
   ...
}
```

Chyroc:
should i add `// Code generated by cmd/generate-std-packages. DO NOT EDIT.` on the head of file

PR #169, opened 2018-03-26
Add LenExpr to the ast walker

floren:
walkExpr gets confused now that len() is part of the language itself instead of a builtin; this fixes the problem.


mattn:
Could you please add test?

floren:
PTAL, I have added a test that looks for lenexpr

mattn:
merged https://github.com/mattn/anko/pull/171

PR #181, opened 2018-03-30
Convert types for sending to chan

MichaelS11:


mattn:
The amount of tests you write is always amazing. Thanks.

MichaelS11:
Thank you very much :)

PR #183, opened 2018-03-31
Add ENV & VM delete

MichaelS11:


vm/vmExpr.go
```diff
@@ -713,33 +713,50 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+				return nilValue, env.Delete(whatExpr.String())
 			}
+			if keyExpr.Kind() != reflect.Bool {
+				return nilValue, env.Delete(whatExpr.String())
+			}
+			if keyExpr.Bool() {
+				return nilValue, env.DeleteGlobal(whatExpr.String())
+			} else {
+				return nilValue, env.Delete(whatExpr.String())
+			}
```
Chyroc:
```
+			if e.KeyExpr != nil && keyExpr.Kind() == reflect.Bool && keyExpr.Bool() {
+				return nilValue, env.DeleteGlobal(whatExpr.String())
+			}
+			return nilValue, env.Delete(whatExpr.String())
```

vm/vmExpr.go
```diff
@@ -713,33 +713,50 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+
+		case reflect.Map:
+			if whatExpr.IsNil() {
+				return nilValue, nil
+			}
+			if whatExpr.Type().Key() != keyExpr.Type() {
+				keyExpr, err = convertReflectValueToType(keyExpr, whatExpr.Type().Key())
+				if err != nil {
+					return nilValue, newStringError(e, "cannot use type "+whatExpr.Type().Key().String()+" as type "+keyExpr.Type().String()+" in delete")
```
Chyroc: use `fmt.Sprintf`

MichaelS11: Why is Sprintf faster? 

Sprintf hits the doPrintf code:
https://golang.org/src/fmt/print.go?#L951

I would think that a concat would be faster.

Either way, the performance gain here is most likely minimum since this error statement is often only going to be run one. I tend to like the readability in the concat format better.

vm/vmExpr.go
MichaelS11: Good catch, will fix :)

MichaelS11:
Should be good now

PR #180, opened 2018-03-30
add expr_slice-type

Chyroc:

PR #185, opened 2018-04-02
Add/item in list syntax

Chyroc:


Chyroc:
after https://github.com/mattn/anko/pull/180

mattn:
I wonder this should be expr independently. For example `1 + 1 in [1,2,3]` or `for a in [1,2,3]` will be expected as another expression.

Chyroc:
do you mean
```
expr_slice :
	IDENT '[' expr ':' expr ']'
	{
		$$ = ast.SliceExpr{Value: &ast.IdentExpr{Lit: $1.Lit}, Begin: $3, End: $5}
	}
	| IDENT '[' expr ':' ']'
	{
		$$ = ast.SliceExpr{Value: &ast.IdentExpr{Lit: $1.Lit}, Begin: $3, End: nil}
	}
	| IDENT '[' ':' expr ']'
	{
		$$ = ast.SliceExpr{Value: &ast.IdentExpr{Lit: $1.Lit}, Begin: nil, End: $4}
	}
	| expr '[' expr ':' expr ']'
	{
		$$ = ast.SliceExpr{Value: $1, Begin: $3, End: $5}
	}
	| expr '[' expr ':' ']'
	{
		$$ = ast.SliceExpr{Value: $1, Begin: $3, End: nil}
	}
	| expr '[' ':' expr ']'
	{
		$$ = ast.SliceExpr{Value: $1, Begin: nil, End: $4}
	}

expr_item_in_list :
	expr IN IDENT
	{
		$$ = &ast.IncludeExpr{ItemExpr: $1, ListExpr: ast.SliceExpr{Value: &ast.IdentExpr{Lit: $3.Lit}, Begin: nil, End: nil}}
	}
	| expr IN expr
	{
		$$ = &ast.IncludeExpr{ItemExpr: $1, ListExpr: ast.SliceExpr{Value: $3, Begin: nil, End: nil}}
	}
	| expr IN expr_slice
	{
		$$ = &ast.IncludeExpr{ItemExpr: $1, ListExpr: $3}
	}
```

?

MichaelS11:
What about a slice in a multidimensional slice? Or other complex in statements?

What about a test for something like for `a in 5 in [1, 2, 3} { return a }` ?

Chyroc:
do anko need to support `a in map`?

Chyroc:
@MichaelS11 do you mean `a in 5 in [1, 2, 3]`

MichaelS11:
Yes.

I say do not support `a in map` since we have `a, ok = aMap["key"]`

Chyroc:
You are right, we have indeed implemented a judgment on whether an element is in a map clock。

Then I implemented `item in list`, maybe you can take a look again, then make some comments?

@mattn @MichaelS11

vm/vmExpr.go
```diff
@@ -732,6 +732,28 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+		if err != nil {
+			return nilValue, newError(e.ItemExpr, err)
+		}
+		listExpr, err := invokeExpr(&e.ListExpr, env)
+		if err != nil {
+			return nilValue, newError(&e.ListExpr, err)
+		}
+
+		if listExpr.Kind() != reflect.Slice && listExpr.Kind() != reflect.Array {
+			return nilValue, newStringError(e, "second argument must be slice or array; but get "+listExpr.Kind().String())
```
MichaelS11: Maybe change "but get" to "but have"

vm/vmItemInList_test.go
```diff
@@ -0,0 +1,47 @@
...
+		// for i in list && item in list
+		{script: `list_of_list = [["a"]];for l in list_of_list { return "a" in l }`, runOutput: true},
+
+		// not slice or array
+		// todo: support `"a" in "aaa"` ?
+		{script: `"a" in "aaa"`, runError: fmt.Errorf("second argument must be slice or array; but get string")},
+		{script: `1 in 12345`, runError: fmt.Errorf("type int64 does not support slice operation")},
+
+		// a in item in list
+		{script: `"a" in 5 in [1, 2, 3]`, runError: fmt.Errorf("type bool does not support slice operation")},
```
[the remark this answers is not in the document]
MichaelS11: Also `1++ in [1, 2, 3]` and `1 in 1++`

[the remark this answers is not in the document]
Chyroc: done

Chyroc:
@MichaelS11 

should `1++ in [1, 2, 3]` and `1 in 1++` work? or return `Run error - received: Invalid operation`
```
{script: `1++ in [1, 2, 3]`, runOutput: true},
{script: `3++ in [1, 2, 3]`, runOutput: false},
{script: `1 in 1++`, runOutput: false},
```

vs


```
{script: `1++ in [1, 2, 3]`, runError: fmt.Errorf("Invalid operation")},
{script: `3++ in [1, 2, 3]`,runError: fmt.Errorf("Invalid operation")},
{script: `1 in 1++`, runError: fmt.Errorf("Invalid operation")},
```

MichaelS11:
Testing is much better. Thank you. 

What about a slice in a multidimensional slice? Or other complex in statements?

[1] in [[1], [2], [3]]
[[1]] in [1,2,3]
etc...

Chyroc:
@MichaelS11 add test
```
		{script: `1 in [1]`, runOutput: true},
		{script: `1 in [[1]]`, runOutput: false},
		{script: `1 in [[[1]]]`, runOutput: false},
		{script: `1 in [[1],[[1]],1]`, runOutput: true},
		{script: `[1] in [1]`, runOutput: false},
		{script: `[1] in [[1]]`, runOutput: true},
		{script: `[1] in [[[1]]]`, runOutput: false},
		{script: `[[1]] in [1]`, runOutput: false},
		{script: `[[1]] in [[1]]`, runOutput: false},
		{script: `[[1]] in [[[1]]]`, runOutput: true},
```

MichaelS11:
Great. If wanted, could also add all the tests so far for data on the outside (using input), plus a mixing of the two (VM data plus input data).

Chyroc:
I'm sorry, my English is not very good, so I'll confirm it again.
What you mean is to save the test(`TestItemInList`) as a file in the `t` directory (test the anko code read from the file by the anko binaty) and do the test?
As well as mixing reads from files by anko binary, and calling `vm.Run` directly?

MichaelS11:
For example:

Here is tests with VM maps
https://github.com/mattn/anko/blob/master/vm/vmArraysAndMaps_test.go#L732-L737

Here is tests with outside/Go maps:
https://github.com/mattn/anko/blob/master/vm/vmArraysAndMaps_test.go#L769-L776

Chyroc:
@MichaelS11 equal function in `vm/vm.go` convert string "1" to int 1, so "1" == 1，it is expected?

and should it right:
```
"1" == 1
1 == "1"
"1" in [1]
1 in ["1"]
```
?

Chyroc:
@MichaelS11 add some test with input, need review~

vm/vmExpr.go
```diff
@@ -756,7 +756,23 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+			list := listExpr.Index(i)
+
+			if item.Kind() == reflect.Interface || item.Kind() == reflect.Ptr {
+				item = item.Elem()
+			}
+			if list.Kind() == reflect.Interface || list.Kind() == reflect.Ptr {
+				list = list.Elem()
+			}
+
+			if isNum(item) && isNum(list) {
```
mattn: Just questino. Why only number?

Chyroc: if not this, than 
int(1) != int64(1) when use input, etc...

Chyroc: ok~

MichaelS11:
Could you help me understand why equalNumber is needed? Is there a reason why the equal function is not being used?

Chyroc:
@MichaelS11 

use equal, than 
```
1 == 1.1
1.1 != 1
```
This doesn't work, this is a bug, if it's just that, I can use equal and than fix the equal

but :
```
"1" == 1
"true" == true
```
This seems to be the functional point of equal, but this is not intuition

MichaelS11:
Please use equal then and we can working on addressing any bugs.

I say match JavaScript equal: 
```
console.log(1 == 1.1) // false
console.log(1 != 1.1) // true
console.log("1" == 1) // true
console.log("true" == true) // false
```

Chyroc:
@MichaelS11 i replace equalNumber by equal,

MichaelS11:
LGTM

mattn:
Thank you

Issue #189, opened 2018-04-03
switch support multi case


Chyroc:
1. support `fallthrough`
```
var fmt = import("fmt")

a  = 1

switch a {
case 1:
	fallthrough
case 2:
	fmt.Printf("%d\n",a)
default:
	fmt.Printf("default\n")
}
```

or 

2. support `case a,b:`
```
var fmt = import("fmt")

a  = 1

switch a {
case 1,2:
	fmt.Printf("%d\n",a)
default:
	fmt.Printf("default\n")
}
```

or 

3. support `case a:    case b:`
```
var fmt = import("fmt")

a  = 1

switch a {
case 1:
case 2:
	fmt.Printf("%d\n",a)
default:
	fmt.Printf("default\n")
}
```

Chyroc:
Which program should we support, maybe I can make a pr

MichaelS11:
Matching Go seems good to me.

https://golang.org/ref/spec#Expression_switches

PR #190, opened 2018-04-03
add multiple-case

Chyroc:


Chyroc:
fix https://github.com/mattn/anko/issues/189

vm/vmIfAndSwitch_test.go
```diff
@@ -54,6 +54,29 @@ func TestSwitch(t *testing.T) {
 	os.Setenv("ANKO_DEBUG", "1")
 	tests := []testStruct{
 		// TODO: add tests
+
+		{script: `switch {}`, parseError: fmt.Errorf("syntax error")},
+		{script: `a=1; switch a; {}`, parseError: fmt.Errorf("syntax error")},
+
+		{script: `a=1; switch a {}`, runOutput: int64(1)},
+
+		{script: `a=1; switch a {case 1: return a}`, runOutput: int64(1)},
```
mattn: please add test for `switch a { case: return a}`

Chyroc: ok, and add 
```go
{script: `a=1; switch a {case: return a}`, parseError: fmt.Errorf("syntax error")},

{script: `a=99; switch a {default: return 'default'}`, runOutput: "default"},
```

mattn:
Thank you

PR #191, opened 2018-04-03
Fix/error message lower case string

Chyroc:


mattn:
Thank you. Now anko use lower case strings in error messages. cc @MichaelS11

PR #195, opened 2018-04-04
implement defer statement

mattn:
need tests.


mattn:
Anko's defer statement doesn't work like Go's one since anko doesn't have named return parameters for a function. Anko can not re-write return value. Also current implementation doesn't have recover().

mattn:
```go
func foo() {
  var i = 0
  defer println("foo", i)
  i++
  defer println("bar", i)
  i++
  defer println("bar", i)
  i++
  defer println("bar", i)
  i++
  defer println("bar", i)
}

foo()
```

This works fine.

```
bar 4
bar 3
bar 2
bar 1
foo 0
```

MichaelS11:
Could we put the AnonCallExpr and CallExpr in vmExprFunction.go then be able to use it from` case *ast.DeferStmt` as well as the current places:

```
	case *ast.AnonCallExpr:
		return anonCallExpr(e, env)

	case *ast.CallExpr:
		return callExpr(e, env)
```

Would be really nice if there was a way to merge the two AnonCallExpr & CallExpr so that we do not have to maintain two code areas.

mattn:
This implementation is strange since VM run defer even if it is in if/for block. It must be in func block.

```go
var time = import("time")
for {
  defer println("foo")
  time.Sleep(1e9)
}
```
foo must not be appeared.

MichaelS11:
I still wonder if some of this code is not needed. Say they do defer something that is not a function, like `a++`. If that is passed to say anonCallExpr as an example, that will error and say it is not a function. I think you can just simplify syntax to just DEFER expr.

MichaelS11:
Was curious, is it an issue if the code can defer any expression, not just functions?

PR #196, opened 2018-04-04
add comment syntax

Chyroc:


Chyroc:
If you agree, I will add a doc file, and then add the documentation about the annotation.
Anko has a lot of grammar, so it is very difficult to finish all the documents at once. Maybe we can add the documents of the contents we modified in each pr, and then the documents will be more and more abundant.

MichaelS11:
Would rather see example, like:

https://github.com/mattn/anko/blob/master/vm/exampleHelloWorld_test.go
https://github.com/mattn/anko/blob/master/vm/exampleBasicOperators_test.go
etc...

MichaelS11:
Yes, would rather see examples first.

Do we want to remove # as a comment?

As far as the code go, generally looks good, but tricky to see if there might be any parsing errors.

Chyroc:
@mattn Do we want to remove # as a comment?

mattn:
Currently, I don't want to remove.

MichaelS11:
@mattn so go to be merged?

Chyroc:
Is there any idea about this pr?

mattn:
Thanks.

PR #204, opened 2018-04-05
add runtest-option-params

Chyroc:


mattn:
Why did you remove Example? `Example_` code is useful since it can be seen and tried on godoc.

Chyroc:
change this pr for add "run test options params"

mattn:
Thanks.

PR #213, opened 2018-04-10
 Add ank files into core testing

MichaelS11:


mattn:
Do you have plan to move t/* into testdata?

MichaelS11:
Was thinking about it, if you are alright with that? Will probably need to add at least some new tests for main as well.

Was trying to keep PR's small.

mattn:
Thanks.

PR #214, opened 2018-04-11
Refactor anko.go part 1

MichaelS11:
To make testing much easier, refactoring anko.go. Will have a part 2 after this.

### DIFF
--- a/anko.go
+++ b/anko.go
@@ -8,7 +8,6 @@ import (
 	"fmt"
 	"io/ioutil"
 	"os"
-	"path/filepath"
 	"strings"
 
 	"github.com/mattn/anko/core"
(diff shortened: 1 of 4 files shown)


anko.go
```diff
@@ -17,75 +16,81 @@ import (
...
-const version = "0.0.2"
+const version = "0.0.3"
 
 var (
-	fs   = flag.NewFlagSet(os.Args[0], 1)
-	line = fs.String("e", "", "One line of program")
-	v    = fs.Bool("v", false, "Display version")
-
-	istty = isatty.IsTerminal(os.Stdout.Fd())
+	flagExecute *string
```
mattn: please use non-pointer variable and flag.StringVar instead.

anko.go
```diff
@@ -17,75 +16,81 @@ import (
...
-		reader    *bufio.Reader
-		following bool
-		source    string
-	)
-
-	env := vm.NewEnv()
-	interactive := fs.NArg() == 0 && *line == ""
+func setupEnv() {
+	env = vm.NewEnv()
+	if len(*flagExecute) > 0 || flag.NArg() < 1 {
```
mattn: please use flagExecute != "" instead. len(flagExecute) > 0 will be bit slower.

anko.go
```diff
@@ -17,75 +16,81 @@ import (
...
-		ct.ChangeColor(color, bright, ct.None, false)
-		f()
-		ct.ResetColor()
-	} else {
-		f()
-	}
+func main() {
+	parseFlags()
+	setupEnv()
+	runNonInteractive()
```
mattn: check flagExecutable here?

if flagExecute != "" {
  runNonInteractive()
} else {
  runInteractive()
}

anko.go
```diff
@@ -17,75 +16,81 @@ import (
...
+		source = string(sourceBytes)
 	}
 
-	core.Import(env)
-	AddPackageColortext()
-	packages.DefineImport(env)
+	_, err := env.Execute(source)
+	if err != nil {
+		fmt.Println("Execute error:", err)
+		os.Exit(4)
```
mattn: let's return int value from runInteractive/runNonInteractive. and call os.Exit(r) in main

anko.go
MichaelS11: The performance difference is going to be so very small in this case because it is only called once, so I am fine either way. 

If you are really interested, my understanding is len is faster because len is a built in function.

https://groups.google.com/forum/#!topic/golang-nuts/7Ks1iq2s7FA

anko.go
MichaelS11: Will do

anko.go
mattn: Yes, this is micro benchmark.



```
package main

import (
	"strings"
	"testing"
)

func BenchmarkEqEq(b *testing.B) {
	s := []string{"bar", "", strings.Repeat("o", 9999)}
	n := 0
	for i := 0; i < b.N; i++ {
		if s[i%3] != "" {
			n++
		}
	}
}

func BenchmarkLen(b *testing.B) {
	s := []string{"bar", "", strings.Repeat("o", 9999)}
	n := 0
	for i := 0; i < b.N; i++ {
		if len(s[i%3]) > 0 {
			n++
		}
	}
}
```
This is just small thing for my coding style. :)

MichaelS11: Going to change it to !=

That being said, found them to be the same speed.

```
func BenchmarkEqEq(b *testing.B) {
	s := make([]string, 10000)
	for i := 0; i < 10000; i++ {
		s[i] = "a"
	}
	n := 0
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if s[n] != "" {
			n++
		}
		if n > 9999 {
			n = 0
		}
	}
}

func BenchmarkLen(b *testing.B) {
	s := make([]string, 10000)
	for i := 0; i < 10000; i++ {
		s[i] = "a"
	}
	n := 0
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if len(s[n]) > 0 {
			n++
		}
		if n > 9999 {
			n = 0
		}
	}
}
```

```
C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               1.04 ns/op
BenchmarkLen-8          2000000000               1.04 ns/op
PASS
ok      test    4.425s

C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               1.04 ns/op
BenchmarkLen-8          2000000000               1.04 ns/op
PASS
ok      test    4.403s

C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               1.04 ns/op
BenchmarkLen-8          2000000000               1.04 ns/op
PASS
ok      test    4.409s
```

MichaelS11: Oh, also if the string is bigger, same result.

Also can do same test on a single string and take in account caching.

```
func BenchmarkEqEq(b *testing.B) {
	s := "a"
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if s != "" {
		}
	}
}

func BenchmarkLen(b *testing.B) {
	s := "a"
	b.ResetTimer()
	for i := 0; i < b.N; i++ {
		if len(s) > 0 {
		}
	}
}
```

```
C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               0.26 ns/op
BenchmarkLen-8          2000000000               0.26 ns/op
PASS
ok      test    1.138s

C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               0.26 ns/op
BenchmarkLen-8          2000000000               0.26 ns/op
PASS
ok      test    1.143s

C:\>go test -v -bench=. test
goos: windows
goarch: amd64
pkg: test
BenchmarkEqEq-8         2000000000               0.26 ns/op
BenchmarkLen-8          2000000000               0.26 ns/op
PASS
ok      test    1.151s
```

anko.go
MichaelS11: Good idea

anko.go
MichaelS11: Good idea

anko.go
mattn: Compiler might be optimized recently, As far as I remember, Russ or Iant say `!=` is bits faster.

MichaelS11:
Please review again

mattn:
LGTM. Thanks.

PR #215, opened 2018-04-12
Added TestRunInteractive

MichaelS11:
Added quit() to interactive


MichaelS11:
@mattn 

Seems like colortext is not needed as much any more, what do you think about removing that dependency?

MichaelS11:
Sounds good to me. How is this PR?

mattn:
Thank you

PR #217, opened 2018-04-15
Updated anko, removed colortext

MichaelS11:
Added tests to TestVar & TestBasicOperators


MichaelS11:
Was following in anko.go still needed? Not sure i grasped the idea behind it.

MichaelS11:
Never mind, I get following now. Fixing PR.

MichaelS11:
Should be go to go.

MichaelS11:
Any thoughts @mattn ?

mattn:
I did not confirm, can I think that all of the tests in t directory are tested in somewhere else?

mattn:
And I hope to keep feature to output colors in multi-platform. So, how about to add go-colorable and go-isatty?

MichaelS11:
The 12 ank files under t directory have been copied to:
https://github.com/mattn/anko/tree/master/core/testdata

MichaelS11:
Thought it would be nice to remove the colortext dependency from Anko. As of this PR, there are no dependencies besides Go core, which seemed good to me.

Having color in Anko command line did not seem that needed to me. Before it was really needed for testing but now after moving testing into go test, that is not needed any more.

mattn:
On UNIX, output escape sequence are works well. But it's not on Windows. So I hope to add way to do.

MichaelS11:
First off, please keep in mind that I thought we were in agreement about color text being removed from what was mentioned on this PR: https://github.com/mattn/anko/pull/215

I am not following why it is important to keep color text. What is the typical use case for Aniko? Is color text needed for the typical use case?

Plus, you may have noticed, I have been trying to focus on the minimal core of what is needed for scripting language. Would like to see the core of the language defined, tested, documented and a production 1.x version released. Lot left to do to make that happen.

PR #227, opened 2018-05-10
Added Copy() function on vm.Env

c4s4:
I need this function to clone a VM in the [NeON build tool](http://github.com/c4s4/neon). As *vm.Env* fields are not public, I have to fork Anko project to implement this *Copy()* function. This would be great if this was integrated in the Anko project.


MichaelS11:
@c4s4 
Would be great if tests could be added. If you would like details about adding tests, just let me know.

c4s4:
@MichaelS11 
I have added tests for Copy() function.

vm/copy_test.go
```diff
@@ -0,0 +1,28 @@
...
+import (
+	"testing"
+)
+
+func TestCopy(t *testing.T) {
+	env := NewEnv()
+	env.Define("a", "a")
+	copy := env.Copy()
+	if v, e := copy.Get("a"); e != nil || v != "a" {
+		t.Errorf("Copy doesn't retain original values")
```
mattn: please use lower case string in first word if it is not identifier. and below.

MichaelS11:
I think the copy function could go into the env.go and env_test.go files.

vm/copy.go
```diff
@@ -0,0 +1,21 @@
...
+
+import "reflect"
+
+// Copy the state of the virtual machine environment
+func (env *Env) Copy() *Env {
+	b := false
+	copy := Env {
+		env:       make(map[string]reflect.Value),
+		typ:       make(map[string]reflect.Type),
+		parent:    nil,
```
MichaelS11: Should the parent be copied?

vm/copy_test.go
c4s4: Done

vm/copy.go
c4s4: Done

mattn:
LGTM. pass to @MichaelS11 's look.

mitchmieuxplacer:
Regarding recursive copy, we could add a recursive boolean argument telling if copy is recursive.

mattn:
Or add new DeepCopy?

mattn:
So is this Clone not Copy? How do you think?

mitchmieuxplacer:
I can't run tests, they fail on master.

mitchmieuxplacer:
I must say I don't really see a difference between clone and copy...

MichaelS11:
What about name and external variables?

vm/env.go
```diff
@@ -422,3 +422,42 @@ func (e *Env) Execute(src string) (interface{}, error) {
...
+		typ:       make(map[string]reflect.Type),
+		parent:    e.parent,
+		interrupt: &b,
+	}
+	for name, value := range e.env {
+		copy.env[name] = value
+	}
+	for name, typ := range e.typ {
+		copy.typ[name] = typ
+	}
```
MichaelS11: Can you just call copy instead of the above code?

MichaelS11:
What about name and external variables?

c4s4:
Done

vm/env.go
```diff
@@ -422,3 +422,32 @@ func (e *Env) Execute(src string) (interface{}, error) {
 func (e *Env) Run(stmts []ast.Stmt) (interface{}, error) {
 	return Run(stmts, e)
 }
+
+// Copy the state of the virtual machine environment
+func (e *Env) Copy() *Env {
+	b := false
+	copy := Env {
```
MichaelS11: It doesn't really matter, but would be kind of nice if it was in the same order as the struct declare.
type Env struct {
	name      string
	env       map[string]reflect.Value
	typ       map[string]reflect.Type
	parent    *Env
	interrupt *bool
	external  EnvResolver

vm/env.go
```diff
@@ -422,3 +422,32 @@ func (e *Env) Execute(src string) (interface{}, error) {
...
+		interrupt: &b,
+		name:      e.name,
+		external:  e.external,
+	}
+	for name, value := range e.env {
+		copy.env[name] = value
+	}
+	for name, typ := range e.typ {
+		copy.typ[name] = typ
+	}
```
MichaelS11: Should there be a lock when coping the maps?

vm/env.go
c4s4: Reordered.

vm/env.go
c4s4: Added lock on copy.

mattn:
Thank you

mitchmieuxplacer:
Thanks

PR #228, opened 2018-05-12
Added regexp and arrays examples

MichaelS11:
Renamed files


MichaelS11:
Would like to remove _example after testable examples have been created.

mattn:
Thank you

PR #235, opened 2018-05-21
Updated readme and for continue example

MichaelS11:


MichaelS11:
@mattn 

I think we are very close to being able to remove the _example/scripts directory. There are many examples I do not want to add, like exec and env because they can make for more difficult test cases. (Being able to test successfully on both Linux and Windows.) Also do not want to add the try catch example since you have talked about removed that from Anko.

Was there any particular test cases you want added that have not been already?

README.md
```diff
@@ -11,101 +11,70 @@ Anko is a scriptable interpreter written in Go.
...
 ```
 
-# Author
+### Running an Anko script file named script.ank
+```
+./anko script.ank
+```
+or 
+```
+anko.exe script.ank
```
mattn: please remove .exe

vm/example_containers_test.go
```diff
@@ -63,7 +63,7 @@ println(a[1])
 
 	_, err = env.Execute(script)
 	if err != nil {
-		log.Fatalf("execute error: %v\n", err)
+		log.Fatalf("Execute error: %v\n", err)
```
mattn: I prefer leading lower case letter.

MichaelS11: Define error is upper case. The reason for the upper case is because it is matching the function call. This is a log message not a error message returned by a function. A function error message should be lower case.

README.md
MichaelS11: Done

vm/example_containers_test.go
mattn: please look codes in go/cmd/**/*.go, all of `log.Fatal` use lower case message.

MichaelS11: Done

mattn:
LGTM, Thanks!

PR #236, opened 2018-05-21
add LetMapItemStmt to astutil/walk

floren:
Add a test too.


mattn:
I don't make sure but LetMapItemStmt is not passed in this test.
https://codecov.io/gh/mattn/anko/compare/0895bd15eb3d7ca6e506a59c3816d6548a7149ad...fa1ac28582bd67f985eaa6d3ad265593209596a8/src/ast/astutil/walk.go

floren:
it looks like pretty much every type of expr lacks coverage in the same place my new code does. I do not know how to write a test to provide that coverage, but I *do* know that my patch fixes a WalkExpr which was previously failing if it encountered a LetMapItemStmt.

mattn:
Okay, merged.

PR #237, opened 2018-05-22
Removed _example

MichaelS11:


TrueFurby:
Why remove the examples? Are you going to move them to some single .md file?

mattn:
I'm thinking some of them is still useful to read script and try to run easily. For example, server.ank? I'm not strong opinion.

MichaelS11:
The server one should be testable, just a little more tricky. Would have to play with it.

mattn:
I'm thinking that server.ank or http.ank will be able to surprise first-time users. :)

MichaelS11:
Working on it

MichaelS11:
Have regexp already. 

There are two fib, the function one is there already. I personally think the for loops are good as is without having to adding the fib loop example.

Have a good idea how to do env? The current example is not testable.

Not sure if signal will work as testable.

Working on http and server.

What about z-combinator is not already in the function examples?

mattn:
I mean how user can look or try to run examples easily. When user have anko(exe) command, they can do `anko /path/to/script.ank` simply.

MichaelS11:
I guess. Don't think it is that hard to copy and paste examples either.

For me, I usually have coping and pasting examples anyways. I take the parts I want and add them to the code I am working on.

In my option the main reasons for the switch is to have the examples testable and in the documentation.

MichaelS11:
Like I was saying in the other thread, was thinking lead people toward the documentation instead. I have seen it done both ways. I personally like guiding people to the documentation but open to either way.

mattn:
I want to take care of the first-time user. The users who look examples mostly be a Go user.

MichaelS11:
Ok :)

MichaelS11:
Will open new PR later

PR #240, opened 2018-05-24
Added http example

MichaelS11:
Added Duration to time package

### DIFF
--- a/packages/time.go
+++ b/packages/time.go
@@ -1,6 +1,7 @@
 package packages
 
 import (
+	"reflect"
 	"time"
 )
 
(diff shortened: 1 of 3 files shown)


MichaelS11:
@mattn 

This is the code I would like to use for the example:

`client = &http.Client{Timeout: 30 * time.Second}`

Which doesn't work.

So tried to get this code to work:

```
duration = make([]time.Duration)
duration[0] = 30 * time.Second
client = http.DefaultClient
client.Timeout = duration[0]
```
Which gives the error: "execute error: type time.Duration cannot be assigned to type time.Duration for struct"

The main issue is if there is a way to make Anko be able to assign values to a struct?

MichaelS11:
@mattn 

Any thoughts about the struct assign issue above?

packages/time.go
```diff
@@ -28,8 +29,10 @@ func init() {
 		"Minute":          time.Minute,
 		"Hour":            time.Hour,
 	}
+	var duration time.Duration
 	PackageTypes["time"] = map[string]interface{}{
-		"Ticker": time.Ticker{},
-		"Time":   time.Time{},
+		"Duration": reflect.TypeOf(&duration).Elem(),
```
mattn: time.Duration(0) ?

MichaelS11: Updated

mattn:
Thank you

Issue #239, opened 2018-05-24
Map evaluation


c4s4:
This script:

```golang
package main

import (
	"fmt"
	"github.com/mattn/anko/vm"
)

func main() {
	m := map[interface{}]interface{}{
		"foo": 1,
		"bar": 2,
	}
	fmt.Println(m["foo"])
	env := vm.NewEnv()
	err := env.Define("println", fmt.Println)
	if err != nil {
		panic(err)
	}
	err = env.Define("m", m)
	if err != nil {
		panic(err)
	}
	_, err = env.Execute(`println(m["foo"])`)
	if err != nil {
		panic(err)
	}
}
```

Produces following output:

```
$ go run test.go 
1
<nil>
```

I think this should output *1* and *1*.

This works as expected with following code:

```golang
package main

import (
	"fmt"
	"github.com/mattn/anko/vm"
)

func main() {
	m := map[string]int{
		"foo": 1,
		"bar": 2,
	}
	fmt.Println(m["foo"])
	env := vm.NewEnv()
	err := env.Define("println", fmt.Println)
	if err != nil {
		panic(err)
	}
	err = env.Define("m", m)
	if err != nil {
		panic(err)
	}
	_, err = env.Execute(`println(m["foo"])`)
	if err != nil {
		panic(err)
	}
}
```

Thus, I suspect this is a problem with *interface{}* and *string* types as keys.

MichaelS11:
Anko supports maps with strings as keys. It has limited support for maps with other keys.

Also see:
https://github.com/mattn/anko/issues/220

c4s4:
Note that keys **are** strings.

c4s4:
OK, but this one has working example code.

c4s4:
I think these issues are not exactly the same. One is about keys type being string only, the other is about type behind *interface{}* being discovered. In my example, keys are strings and Go is able to get the value because it gets the real type of the key, not the declared one.

c4s4:
I have made a pull request to fix this issue. *Enjoy!*

MichaelS11:
@mattn 
I think this can be closed.

PR #241, opened 2018-05-24
Fixed issue #239: access map of interfaces with string keys

c4s4:


MichaelS11:
Please move test out of TestMakeMapsData and into TestMaps. Can use the Types to pass along the needed map type. Can also add additional tests.

vm/vm.go
```diff
@@ -211,7 +211,7 @@ func getMapIndex(key reflect.Value, aMap reflect.Value) reflect.Value {
 			keyType = key.Elem().Type()
 		}
 	}
-	if keyType != aMap.Type().Key() {
+	if keyType != aMap.Type().Key() && aMap.Type().Key() != interfaceType {
```
MichaelS11: This should change use convertReflectValueToType at some point.

c4s4: I can't move test from *TestMakeMapsData()* to *TestMaps()* because if we set map in an Anko script, keys would be *string*, not *interface{}*.

c4s4: Added more unit tests.

MichaelS11:
Can create a custom map type:

{Script: `make(mapInterfaceString)`, Types: map[string]interface{}{"mapInterfaceString": map[interface{}]stringl{}}, ....

c4s4:
I have added  a test in *TestMaps()* but it doesn't work while it should.

c4s4:
Done

MichaelS11:
Please remove tests out of TestMakeMapsData. TestMakeMapsData is handling some very particular cases.

Also please reorder tests, Types is between Script and Output. (Script, ParseError, Types, Input, RunOutput, Output)

Instead of doing == and testing RunOutput for true and false, can you please just check the value?

So the first three tests would look something like these two tests (can remove the extra one):
```
{Script: `a = make(mapInterfaceInt64); a["b"] = 1; a["b"]`, Types: map[string]interface{}{"mapInterfaceInt64": map[interface{}]int64{}}, RunOutput: int64(1), Output: map[string]interface{}{"a": map[interface{}]int64{"b": int64(1)}}},
{Script: `a = make(mapInterfaceInt64); a.b = 1; a.b`, Types: map[string]interface{}{"mapInterfaceInt64": map[interface{}]int64{}}, RunOutput: int64(1), Output: map[string]interface{}{"a": map[interface{}]int64{"b": int64(1)}}},
```

Can you add some tests for void/not found? Like: `{Script: `a = make(mapInterfaceInt64); a["b"] = 1; a["c"]``...

Can you add some additional tests using Input instead of Types and make?

mitchmieuxplacer:
Done

MichaelS11:
LGTM

mattn:
Thank you

c4s4:
Thanks

PR #244, opened 2018-05-31
Removed _examples

MichaelS11:


mattn:
I don't mind that some examples doesn't have tests. As I mentioned in https://github.com/mattn/anko/pull/237, please keep some scripts, or put it on README.md

MichaelS11:
I don't think that readme is the right place to put a lot of examples. In my option the readme is how to get people started and important notes. The documentation is a great place to put the details.

If you don't mind that some examples do not have good tests, we should be able to add the rest of the examples to the documentation. 

So that should leave the question to, what is left to add? From https://github.com/mattn/anko/pull/237 I think just signal and env, is that correct? Believe the rest have been added in some form or another. (server.ank is in the http example)

mattn:
Please see https://github.com/mattn/anko/pull/237#issuecomment-391780728 again

I want to take care of the first-time user. The users who look examples mostly be a Go user (like you).

MichaelS11:
I agree that we both want to take care of the first time user as well as everyone else. Don't think that is in question. I also think we both want examples. Just trying to work out what examples and where to put them.

My biggest option is to keep the readme simple. How it is now is great to me. :) 

So I think we either should either move the wanted examples into the documentation or keep _examples directory. I prefer moving the examples into the documentation.

If you are alright with moving the example into the documentation, I just need to understand clearly what examples are left to move. From my understanding only singal and env are left. Or was there more?

mattn:
How about this?

chan.ank
env.ank
exec.ank
fib.ank
signal.ank
z-combinator.ank

MichaelS11:
Thank you. I think this is where I am having issues following. Let's just look at just **chan.ank** for the moment. There the channel example in the documentation here:
https://godoc.org/github.com/mattn/anko/vm#example-package--VmChannels
Which is located in the code here:
https://github.com/mattn/anko/blob/master/vm/example_operators_test.go#L348-L379

I thought that covered everything that is in the chan.ank file. Do you think something is missing in the documentation example? Could you help me better understand what you are asking for?

mattn:
I'm okay that some examples are duplicated in _example and example_test. So I want to keep it on.

MichaelS11:
Definitely confused now. Let's back up and start fresh. I am trying to remove _example/scripts directory. After we move out the wanted examples to the documentation, can I remove the _example/scripts directory?

mattn:
> can I remove the _example/scripts directory?

No.

PR #245, opened 2018-05-31
Fix case

MichaelS11:


mattn:
Thank you. I was going to fix it.

PR #248, opened 2018-05-31
Nil coalescing operator

earncef:
Added 'if invalid' operator similar to Swift's nil coalescing operator, adapted for scripting. This can be extremely useful in eliminating unnecessarily lengthy code.

The difference being that instead of checking for nil the operator checks to see if the left side is valid (error, empty slice, empty map or zero value).


mattn:
Thanks Sounds good to me.

@MichaelS11 How do you think?

MichaelS11:
Looking at https://en.wikipedia.org/wiki/Null_coalescing_operator I noticed it is done a few different ways. Sometime they do not remove errors, only handle nulls. 

The way this is being implemented here, it removes errors. For example if a variable is not found, it would normally error, but here it is treated as a null. Is that what we want? I can see both ways as an option, just not sure if it is a good idea to hide the errors.

vm/vmExpr.go
```diff
@@ -560,6 +560,19 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
 		}
 		return rhsV, nil
 
+	case *ast.IfInvalidOpExpr:
+		lhsV, err := invokeExpr(e.Lhs, env)
+		isZeroValue := reflect.DeepEqual(lhsV.Interface(), reflect.Zero(lhsV.Type()).Interface())
```
MichaelS11: Can this be done without using DeepEqual?

vm/vmExpr.go
```diff
@@ -560,6 +560,19 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+		isEmpty := (lhsV.Kind() == reflect.Slice || lhsV.Kind() == reflect.Map) && lhsV.Len() == 0
```
MichaelS11: isEmpty really nil?

earncef: It's not in Go, but from a scripting perspective, it's usually the case that an empty slice or a map can be checked in scripting languages with the equivalent of:
if !variable {
}

vm/vmExpr.go
earncef: I'll have to look into this.

earncef:
As for handling errors, I'm not sure this would be the ideal way to do it but I think there is a case to be made for both ways.

A scenario where this is very useful is slices and maps. For instance, while dealing with json it can be very useful to check something of the form a['b']['c'] without having to check for a['b'] but that would throw an error at the moment. 

But at the same time a missing variable is a scenario where it might be better to return an error.

earncef:
I was looking for tests of the ternary operator In order to get a better view on how anko currently deals with the philosophy of the nil value of a slice but there seem to be no tests for the ternary operator.

So I looked into the tests for toBool() method which is used by the ternary operator and it doesn't seem to handle slices or maps and considers non empty strings as an error. This means that the ternary operator works in strange and unexpected ways for expressions like these:
b = [2, 3]; a = b ? 3: 1 // A is set to 1 irrespective of what b is
b = [2, 3]; a = b[1:2] ? 3: 1 // A is set to 1 irrespective of what b is
b = "something"; a = b ? 3: 1 // A is set to 1 irrespective of what b is (unless b is "true")

These expressions make no sense because while toBool calls tryToBool which returns an error when it can't handle a certain type or for invalid strings, the ternary operator doesn't throw an error in such invalid cases but goes along as if the statements were false, which is nonintuitive. The same goes for if/else statements. I don't think this is the best way to do it. I also see no tests for the above described scenarios, so I'm guessing they haven't been thought out yet.

Every dynamically typed language I've looked into seems to have a true/false value for expressions like this depending on the value of the variable and not a blanket false because the variable was an incompatible type. 

I take it that the philosophy with toBool() in anko is to return the statically typed equivalent of bool and I think this should be left as is. But I think the true/false value of an expression might need some thought.

Thoughts?

mattn:
please add test to make sure RHS is not evaluated when LHS is true. For example

```
1 ?? panic(2)
```

vm/vmExpr.go
```diff
@@ -780,3 +791,44 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+	case reflect.Int64, reflect.Int32, reflect.Int16, reflect.Int8, reflect.Int:
+		return v.Int() != 0
+	case reflect.Bool:
+		return v.Bool()
+	case reflect.String:
+		s := v.String()
+		if v.Len() == 0 {
+			return false
+		}
+		if s == "true" {
```
MichaelS11: For true, can it be any of the following: 1, t, T, TRUE, true, True
Like ParseBool:
https://golang.org/pkg/strconv/#ParseBool

MichaelS11:
Do we want both both tryToBool and exprBoolValue?

earncef:
I'll run some tests merging exprBoolValue into tryToBool.

earncef:
I've replaced the old tryToBool with exprBoolValue along with changes to evaluate using parseBool. The case sensitive flag in the old tryToBool is no longer relevant and is removed.

vm/vmToX.go
```diff
@@ -51,18 +49,23 @@ func tryToBool(v reflect.Value, caseSensitive bool) (bool, error) {
 		return v.Bool(), nil
 	case reflect.String:
 		s := v.String()
-		if !caseSensitive {
-			s = strings.ToLower(s)
+		if v.Len() == 0 {
```
MichaelS11: I think the reflect.String can be simplified to just use ParseBool, remove all the extra stuff before and after it. Also please check the error from ParseBool and return if needed.

MichaelS11:
I see some naming differences, I see IfInvalidOpExpr and TestNilCoalescingOperator. Could it be called Nil Coalescing Operator only?

ast/expr.go
```diff
@@ -88,6 +88,14 @@ type BinOpExpr struct {
 	Rhs      Expr
 }
 
+// IfInvalidOpExpr provide if invalid operator expression.
+type IfInvalidOpExpr struct {
+	ExprImpl
+	Expr Expr
```
MichaelS11: Is Expr needed?

MichaelS11:
Thanks for adding the new tests to TestComparisonOperators, awesome! :)

vm/vmToX.go
earncef: ParseBool doesn’t work on “0.0” and also isn’t enough to validate whether the string is nonempty. An error from ParseBool means that the contents of the string couldn’t be converted to bool by Go specification but the string itself might still be non empty, contain a float, an integer not 0 or 1, etc. These errors aren’t really relevant to tryToBool which tries to get the Boolean value of the string itself rather than just its contents.

mattn:
LGTM?

vm/vmToX.go
MichaelS11: Good point.
Not a big deal, but I think v.Len() == 0 can be moved up before s := v.String().

MichaelS11:
LGTM

mattn:
Thank you

PR #258, opened 2018-06-23
More enabling map of any type & If scope

MichaelS11:


MichaelS11:
Updated IfStmt scope and added scope tests

mattn:
Thank you

Issue #262, opened 2018-07-09
Question： range(m, n)


catplant:
why range(m, n) returns [m, m+1, ..., n-1, n] while range(n) returns [0, 1, ..., n-1]?
I think range(n) should be equal to range(0, n)

mattn:
@MichaelS11 How do you think?

vim: range(1) => [0], range(1,3) => [1,2,3]
python: range(1) => [0], range(1,3) => [1,2]

MichaelS11:
I think like Python is good:
range(stop)
range(start, stop[, step])

https://docs.python.org/3/library/functions.html#func-range

https://stackoverflow.com/questions/3895478/does-javascript-have-a-method-like-range-to-generate-a-range-within-the-supp

At least for the short term. Long term, probably should change how range works, at least internally. Making a range of say 0 to 1,000,000,000,000,000,000 would make a really big array.

PR #263, opened 2018-07-11
fix https://github.com/mattn/anko/issues/262 : range

Chyroc:


core/core.go
```diff
@@ -27,23 +27,30 @@ func Import(env *vm.Env) *vm.Env {
...
-			min = args[0]
-			max = args[1]
+		var start, stop int64
+		var step int64 = 1
+
+		switch argsLen := len(args); argsLen {
+		case 0:
+			panic("range expected 1 arguments, got 0")
+		case 1:
+			start = 0
```
MichaelS11: No big deal but start is already 0

core/core.go
```diff
@@ -27,23 +27,30 @@ func Import(env *vm.Env) *vm.Env {
...
-			max = args[0] - 1
-		} else {
-			min = args[0]
-			max = args[1]
+		var start, stop int64
+		var step int64 = 1
+
+		switch argsLen := len(args); argsLen {
+		case 0:
+			panic("range expected 1 arguments, got 0")
```
MichaelS11: Change to something like: range needs at least 1 argument, got 0

Chyroc: remove it

core/core.go
Chyroc: remove it

core/core.go
Chyroc: done

core/core.go
```diff
@@ -27,23 +27,29 @@ func Import(env *vm.Env) *vm.Env {
...
-			max = args[0] - 1
-		} else {
-			min = args[0]
-			max = args[1]
+		var start, stop int64
+		var step int64 = 1
+
+		switch argsLen := len(args); argsLen {
+		case 0:
+			panic("range expected at least 1 arguments, got 0")
```
MichaelS11: arguments -> argument

MichaelS11:
Could you add some more tests with 0 in them?
For example range(x,0,x)
For example range(x,x,0)

MichaelS11:
For reference, Interesting way JavaScript does it:

https://github.com/lodash/lodash/blob/master/.internal/baseRange.js

core/core.go
```diff
@@ -27,23 +27,32 @@ func Import(env *vm.Env) *vm.Env {
...
+			}
+		default:
+			panic(fmt.Sprintf("range expected at most 3 arguments, got %d", len(args)))
 		}
+
 		arr := []int64{}
-		for i := min; i <= max; i++ {
+		for i := start; (step > 0 && i < stop) || (step < 0 && i > stop); {
 			arr = append(arr, i)
+			i += step
```
MichaelS11: Move into for loop?

MichaelS11:
LGTM

mattn:
Thank you

PR #268, opened 2018-07-22
fix switch variable scope

Chyroc:


MichaelS11:
From my understanding, each case or default statement needs it's own new environments under the switch new environment.

Also, could you add some more tests? Make sure anything under the switch, case, or default environments are gone after the switch. That anything in the switch environment is usable under the case or default. Plus to make sure anything under a case environment is not usable under another case or the default environment.

MichaelS11:
Was looking for a test case something like

a=1; switch b = 2; b { case 2: a = 2; }; a

MichaelS11:
LGTM

MichaelS11:
Was doing some parser work and switch got pulled into what I was working on. Ending up redoing how switch worked, so the changes in this PR were manually pulled into that work:

https://github.com/mattn/anko/pull/271

This PR should be able to be closed, if the other PR is approved.

MichaelS11:
This can closed now because of PR https://github.com/mattn/anko/pull/271

PR #271, opened 2018-07-26
Changed CForStmt. Updated parser for ident, var, lets, and switch

MichaelS11:
This has the changes from https://github.com/mattn/anko/pull/268 in it.
Did some rework of switch, so that PR would conflict.


MichaelS11:
This is a first pass on fixing some of the parsing of var/let in the for loop. Got side tracked into the switch code.

Thought it would be good to merge before too much code change.

MichaelS11:
Fix when map item is not found it is not equal to nil and should be.
https://github.com/mattn/anko/issues/272

mattn:
Thank you

PR #283, opened 2018-12-14
fix convert value

catplant:
fix convert value

### DIFF
--- a/vm/vmContainers_test.go
+++ b/vm/vmContainers_test.go
@@ -1578,7 +1578,26 @@ func TestStructs(t *testing.T) {
 				A interface{}
 				B interface{}
 			}{A: int64(1), B: int64(3)}}},
-
+		{Script: `a.A = 1.2; a.B = 1.3;`, Input: map[string]interface{}{"a": &struct {
+			A float64
+			B int64
+		}{A: float64(1), B: int64(2)}},
+			RunOutput: float64(1.3),
+			Output: map[string]interface{}{"a": &struct {
+				A float64
+				B int64
+			}{A: float64(1.2), B: int64(1)}}},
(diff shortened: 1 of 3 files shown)


vm/vmLetExpr.go
```diff
@@ -56,19 +56,12 @@ func invokeLetExpr(expr ast.Expr, rv reflect.Value, env *Env) (reflect.Value, er
...
 			}
 
-			if v.Kind() == reflect.Func && rv.Kind() == reflect.Func {
-				rv, err = convertReflectValueToType(rv, v.Type())
-				if err == nil {
-					v.Set(rv)
-					return v, nil
-				}
-			}
-
```
MichaelS11: Would change to

```
			rv, err = convertReflectValueToType(rv, v.Type())
			if err != nil {
				return nilValue, newStringError(expr, "type "+rv.Type().String()+" cannot be assigned to type "+v.Type().String()+" for struct")
			}

			v.Set(rv)
			return v, nil
```

vm/vmContainers_test.go
```diff
@@ -1578,7 +1578,26 @@ func TestStructs(t *testing.T) {
 				A interface{}
 				B interface{}
 			}{A: int64(1), B: int64(3)}}},
-
+		{Script: `a.A = 1.2; a.B = 1.3;`, Input: map[string]interface{}{"a": &struct {
```
MichaelS11: Test a.A = 1.2; a.B = 1.3 does not seem to test anything new. It also does not seem to add any coverage.

MichaelS11:
Thank you very much for finding this issue and coming up with a fix.

There was a few changes in the fix so instead of going over all the changes just made a new PR.

https://github.com/mattn/anko/pull/284

MichaelS11:
Think this PR can now be closed

PR #284, opened 2018-12-18
Convert interface nil fix

MichaelS11:


MichaelS11:
@catplant Thanks for the report and PR https://github.com/mattn/anko/pull/283

MichaelS11:
@mattn Good to go

mattn:
Thanks

PR #285, opened 2018-12-19
Add context so we can interrupt channels

alaingilbert:
**Problem:** When you have a script that ends with "receive from chan", the script get stuck there forever, and `vm.Interrupt(env)` is unable to cancel the execution of that script.
Example:
```
ch = make(chan string)
<-ch
```
**Solution:** I added an optional `(e *Env) SetContext(context.Context) *Env` function.
When you give a context to this function, it will change how it interrupt the code inside the VM.
There is 2 main differences
- The main loop will check for each statement if there is a context, and if it should exit the loop.
- The channels will now "select" between the `ctx.Done()` and the channel, instead of just waiting forever on the `ch.Recv()`

If there is no context, none of the above apply, the code execute as it was before.

The test [TestInterruptChannelWithContext](https://github.com/mattn/anko/pull/285/files#diff-d11f06e0aa37028f414640b3ea88c2acR827) will break the current master

### DIFF
--- a/vm/vmExpr.go
+++ b/vm/vmExpr.go
@@ -736,7 +736,18 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 					if err != nil {
 						return nilValue, newStringError(e, "cannot use type "+rhs.Type().String()+" as type "+chanType.String()+" to send to chan")
 					}
-					lhs.Send(rhs)
+					cases := []reflect.SelectCase{{
+						Dir:  reflect.SelectRecv,
+						Chan: reflect.ValueOf(ctx.Done()),
+						Send: reflect.ValueOf(nil),
+					}, {
+						Dir:  reflect.SelectSend,
+						Chan: lhs,
+						Send: rhs,
+					}}
+					if chosen, _, _ := reflect.Select(cases); chosen == 0 {
+						return nilValue, ErrInterrupt
+					}
 				}
 				return nilValue, nil
 			} else if rhs.Kind() == reflect.Chan {
--- a/vm/vmExpr.go
+++ b/vm/vmExpr.go
@@ -710,11 +710,11 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 				cases := []reflect.SelectCase{{
 					Dir:  reflect.SelectRecv,
 					Chan: reflect.ValueOf(ctx.Done()),
-					Send: reflect.ValueOf(nil),
+					Send: zeroValue,
 				}, {
 					Dir:  reflect.SelectRecv,
 					Chan: rhs,
-					Send: reflect.ValueOf(nil),
+					Send: zeroValue,
 				}}
 				chosen, rv, _ := reflect.Select(cases)
 				if chosen == 0 {
@@ -739,7 +739,7 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 					cases := []reflect.SelectCase{{
 						Dir:  reflect.SelectRecv,
 						Chan: reflect.ValueOf(ctx.Done()),
-						Send: reflect.ValueOf(nil),
+						Send: zeroValue,
 					}, {
 						Dir:  reflect.SelectSend,
 						Chan: lhs,
@@ -754,11 +754,11 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 				cases := []reflect.SelectCase{{
 					Dir:  reflect.SelectRecv,
 					Chan: reflect.ValueOf(ctx.Done()),
-					Send: reflect.ValueOf(nil),
+					Send: zeroValue,
 				}, {
 					Dir:  reflect.SelectRecv,
 					Chan: rhs,
-					Send: reflect.ValueOf(nil),
+					Send: zeroValue,
 				}}
 				chosen, rv, ok := reflect.Select(cases)
 				if chosen == 0 {
--- a/vm/vmExpr.go
+++ b/vm/vmExpr.go
@@ -707,7 +707,6 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 
 		if e.Lhs == nil {
 			if rhs.Kind() == reflect.Chan {
-				var rv reflect.Value
 				cases := []reflect.SelectCase{{
 					Dir:  reflect.SelectRecv,
 					Chan: reflect.ValueOf(ctx.Done()),
@@ -717,8 +716,7 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 					Chan: rhs,
 					Send: reflect.ValueOf(nil),
 				}}
-				var chosen int
-				chosen, rv, _ = reflect.Select(cases)
+				chosen, rv, _ := reflect.Select(cases)
 				if chosen == 0 {
 					return nilValue, ErrInterrupt
 				}
@@ -742,8 +740,6 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 				}
 				return nilValue, nil
 			} else if rhs.Kind() == reflect.Chan {
-				var rv reflect.Value
-				var ok bool
 				cases := []reflect.SelectCase{{
 					Dir:  reflect.SelectRecv,
 					Chan: reflect.ValueOf(ctx.Done()),
@@ -753,8 +749,7 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 					Chan: rhs,
 					Send: reflect.ValueOf(nil),
 				}}
-				var chosen int
-				chosen, rv, ok = reflect.Select(cases)
+				chosen, rv, ok := reflect.Select(cases)
 				if chosen == 0 {
 					return nilValue, ErrInterrupt
 				}
--- a/vm/vmExpr.go
+++ b/vm/vmExpr.go
@@ -730,7 +730,18 @@ func invokeExpr(expr ast.Expr, env *Env, ctx context.Context) (reflect.Value, er
 			if lhs.Kind() == reflect.Chan {
 				chanType := lhs.Type().Elem()
 				if chanType == interfaceType || (rhs.IsValid() && rhs.Type() == chanType) {
-					lhs.Send(rhs)
+					cases := []reflect.SelectCase{{
+						Dir:  reflect.SelectRecv,
+						Chan: reflect.ValueOf(ctx.Done()),
+						Send: zeroValue,
+					}, {
+						Dir:  reflect.SelectSend,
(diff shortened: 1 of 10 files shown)


mattn:
I prefer RunSingleStmtContext and RunStmtContext.

MichaelS11:
Yeah, I agree, remove SetContext and instead add ExecuteContext, RunContext in env. Like Matt said, for vm add RunSingleStmtContext. For the other one, shouldn't it just be RunContext instead of RunStmtContext? I don't think a RunStmtContext is needed, right?

Also, seems like it would be a good idea to make this a breaking change and remove interrupt. Having two ways to stop the running seems unneeded.

alaingilbert:
@MichaelS11 I'll edit this message later to address the other points.
I think the reason to not got with a channel all the way was to avoid having the "select{}" in the main loop since it is much slower than checking the bool. (https://github.com/mattn/anko/pull/20)
Which I think it's true, not the first time I hear that.
And that's why I made it optional. So if you don't use the context, you don't get the penalty.
That being said, In the end, you have the final word !

MichaelS11:
Yes, if is faster than select but the performance difference is quite small, do not think that should be the main consideration. Simple benchmark:

BenchmarkSelect-8       200000000                7.58 ns/op
BenchmarkIf-8           1000000000               2.36 ns/op

There are many parts of the code that will be much much slower than the if vs select.

Having only context simplifies the code and the usage of the code. Execute as an example, will just call ExecuteContext with context.Background() which means everything will really run from ExecuteContext. This means not having two main code paths or ways of doing things.


Matt has the final word ;)

alaingilbert:
Alright, I'll refactor my PR tonight and simplify the whole thing :)

MichaelS11:
My thought was the context would not be saved inside the env so that if two or more programs were running and one context was done it would not affect the other programs.

alaingilbert:
I'm pretty sure that each env
```go
env := vm.NewEnv().SetContext(ctx)
```
have their own closure.
I did some tests, and if I starts many "NewEnv" at the same time, I can cancel them individually.

alaingilbert:
Oh wow, hum ok, I kinda fail to see ause case for this.
but I think you could just change the ctx for the child env.
eg:
```go
ctxParent, cancelParent := context.WithCancel(context.Background())
envParent := NewEnv().SetContext(ctxParent)

ctxChild, cancelChild := context.WithCancel(context.Background())
envChild := envParent.NewEnv().SetContext(ctxChild)
```

MichaelS11:
Think SetContext should be removed completely. 

The context can just be passed from the ExecuteContext, RunContext, or RunContext to all the child functions. There will be a lot of function to touch/change but in the long run I think it will work better since it will not have to depend on the env.

alaingilbert:
Ok, I think I understand what you guys are talking about.

```go
env := vm.NewEnv()
if err := env.Execute(script1); err != nil {...}
if err := env.Execute(script2); err != nil {...}
if err := env.Execute(script3); err != nil {...}
```
You would prefer if it was possible to just cancel the execution of `script2` for example.
If I understand properly.

Something like this:
```go
env := vm.NewEnv()
if err := env.Execute(script1); err != nil {...}
if err := env.ExecuteContext(script2, ctx); err != nil {...}
if err := env.Execute(script3); err != nil {...}
```

MichaelS11:
Yes :)

MichaelS11:
As I mentioned earlier, the normal function, say Execute, should call ExecuteContext with context.Background(). Then all the `if ctx != nil {` can be removed.

alaingilbert:
Interesting, I missed that part. Let me fix this.

vm/vmStmt.go
```diff
@@ -20,25 +26,52 @@ func Run(stmts []ast.Stmt, env *Env) (interface{}, error) {
...
-		switch stmt.(type) {
-		case *ast.BreakStmt:
-			return nilValue, ErrBreak
-		case *ast.ContinueStmt:
-			return nilValue, ErrContinue
-		case *ast.ReturnStmt:
-			rv, err = runSingleStmt(stmt, env)
-			if err != nil {
-				return rv, err
+	if ctx != nil {
```
MichaelS11: This if else block can be removed.

alaingilbert:
Alright, I fixed it. Thanks for the good advice !

vm/vm_test.go
```diff
@@ -778,41 +780,69 @@ func runInterruptTest(t *testing.T, script string) {
...
 	go func() {
 		<-waitChan
-		Interrupt(env)
+		cancel()
 	}()
 
-	_, err = env.Execute(script)
-	if err == nil || err.Error() != ErrInterrupt.Error() {
+	_, err = env.ExecuteContext(script, ctx)
+	if err == nil || !strings.Contains(err.Error(), ErrInterrupt.Error()) {
```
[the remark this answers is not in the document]
alaingilbert: This one is in strings.Contains because the `run` [function](https://github.com/mattn/anko/blob/ad2f2da49381990329902ee5a96f0833ad1cb0e1/vm/vmStmt.go#L29) is [recursive](https://github.com/mattn/anko/blob/ad2f2da49381990329902ee5a96f0833ad1cb0e1/vm/vmStmt.go#L209), and [wrap](https://github.com/mattn/anko/blob/ad2f2da49381990329902ee5a96f0833ad1cb0e1/vm/vmStmt.go#L86) the returning errors.
Which means that the ErrInterrupt error that you receive is [sometime wrapped](https://github.com/mattn/anko/blob/ad2f2da49381990329902ee5a96f0833ad1cb0e1/vm/vmExpr.go#L723), therefore cannot be check with `==`.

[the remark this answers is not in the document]
MichaelS11: That might be fixed once the other changes have been done. If not, we probably need look into it and fix that issue.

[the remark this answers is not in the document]
alaingilbert: Nevermind, I think it was failing when I was trying to compare the error itself (not their message).
The original error checking works just fine.

MichaelS11:
Also could you please add testing that will hit both code:
```
if chosen == 0 {
                                        return nilValue, ErrInterrupt
```

vm/vmExpr.go
```diff
@@ -699,18 +700,32 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+					Dir:  reflect.SelectRecv,
+					Chan: reflect.ValueOf(ctx.Done()),
+					Send: reflect.ValueOf(nil),
+				}, {
+					Dir:  reflect.SelectRecv,
+					Chan: rhs,
+					Send: reflect.ValueOf(nil),
+				}}
+				var chosen int
+				chosen, rv, _ = reflect.Select(cases)
```
MichaelS11: Could this be changed to `chosen, rv, _ := reflect.Select(cases)` ?

vm/vmExpr.go
```diff
@@ -727,33 +742,48 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
...
+					Dir:  reflect.SelectRecv,
+					Chan: reflect.ValueOf(ctx.Done()),
+					Send: reflect.ValueOf(nil),
+				}, {
+					Dir:  reflect.SelectRecv,
+					Chan: rhs,
+					Send: reflect.ValueOf(nil),
+				}}
+				var chosen int
+				chosen, rv, ok = reflect.Select(cases)
```
MichaelS11: Could this be changed to `chosen, rv, ok := reflect.Select(cases)`?

vm/vmStmt.go
```diff
@@ -345,18 +364,30 @@ func runSingleStmt(stmt ast.Stmt, env *Env) (reflect.Value, error) {
...
+					Dir:  reflect.SelectRecv,
+					Chan: reflect.ValueOf(ctx.Done()),
+					Send: reflect.ValueOf(nil),
+				}, {
+					Dir:  reflect.SelectRecv,
+					Chan: val,
+					Send: reflect.ValueOf(nil),
+				}}
+				var chosen int
+				chosen, iv, ok = reflect.Select(cases)
```
MichaelS11: Could this be changed to chosen, iv, ok := reflect.Select(cases)

vm/vmExpr.go
alaingilbert: Now that we don't have "interrupt", yes !

alaingilbert:
I'm not sure where I could add a test to reach the select branches.
Any ideas ?

vm/vm_test.go
```diff
@@ -778,41 +779,67 @@ func runInterruptTest(t *testing.T, script string) {
...
 
-	_, err = env.Execute(script)
+	_, err = env.ExecuteContext(script, ctx)
 	if err == nil || err.Error() != ErrInterrupt.Error() {
 		t.Errorf("execute error - received %#v - expected: %#v", err, ErrInterrupt)
 	}
 }
 
-func TestInterruptConcurrency(t *testing.T) {
+func TestCancelChannelWithContext(t *testing.T) {
```
MichaelS11: Please look closely at where TestCancelChannelWithContext is getting interrupted at. I think using `time.Sleep(time.Millisecond)` is not going to get consistent results. If want to cancel at the correct time (at the channel wait) going to need to use a more complicated script. Look at TestCancelWithContext as to some hints on how.

alaingilbert: Updated

vm/vmExpr.go
```diff
@@ -723,37 +736,60 @@ func invokeExpr(expr ast.Expr, env *Env) (reflect.Value, error) {
 					if err != nil {
 						return nilValue, newStringError(e, "cannot use type "+rhs.Type().String()+" as type "+chanType.String()+" to send to chan")
 					}
-					lhs.Send(rhs)
+					cases := []reflect.SelectCase{{
+						Dir:  reflect.SelectRecv,
+						Chan: reflect.ValueOf(ctx.Done()),
+						Send: reflect.ValueOf(nil),
```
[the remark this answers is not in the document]
MichaelS11: Also a few lines above there seems to be another `lhs.Send(rhs)`, convert that to the select case with ctx?

vm/vmStmt.go
```diff
@@ -345,18 +364,27 @@ func runSingleStmt(stmt ast.Stmt, env *Env) (reflect.Value, error) {
...
+					Dir:  reflect.SelectRecv,
+					Chan: reflect.ValueOf(ctx.Done()),
+					Send: reflect.ValueOf(nil),
+				}, {
+					Dir:  reflect.SelectRecv,
+					Chan: val,
+					Send: reflect.ValueOf(nil),
+				}}
+				chosen, iv, ok := reflect.Select(cases)
+				if chosen == 0 {
```
MichaelS11: Noticed this one is not being covered by the tests. Not required but would be nice if it could be covered by the tests.

alaingilbert: Yeah, I noticed it as well. Couldn't figure out out to reach this part of the software :/

alaingilbert: Updated.

MichaelS11:
@alaingilbert Great work 🥇 

@mattn Looks good to me

mattn:
Thanks

PR #286, opened 2018-12-25
select statement

alaingilbert:
Add `select` to the language

examples:
```go
a = make(chan int64, 1)
select {
    case <-a:
        println("value received on a")
    default:
        println("default")
}
```

```go
a = make(chan int64, 1)
b = make(chan int64, 1)
b <- 2
val = 0
select {
    case val = <-a:
        println("value " + val + " received on a")
    case val = <-b:
        println("value " + val + " received on b")
}
```


vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
 		env.defineGlobalValue(stmt.Name, reflect.ValueOf(newenv))
 		return rv, nil
 
+	// SelectStmt
+	case *ast.SelectStmt:
+		newenv := env.NewEnv()
+		body := stmt.Body.(*ast.SelectBodyStmt)
+		letsStmts := []*ast.LetsStmt{nil}
```
MichaelS11: Is `letsStmts` or `bodies` really needed? Can one use `chosen` later to access wanted LetsStmt and Stmt?

vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
...
+		}}
+		for _, selectCaseStmt := range body.Cases {
+			caseStmt := selectCaseStmt.(*ast.SelectCaseStmt)
+			switch e := caseStmt.Expr.(type) {
+			case *ast.LetsStmt:
+				ee, ok := e.Rhss[0].(*ast.ChanExpr)
+				if !ok {
+					return nilValue, newStringError(e.Rhss[0], "invalid operation")
+				}
+				ident, _ := ee.Rhs.(*ast.IdentExpr)
```
MichaelS11: Move this line and lines after outside of switch statement for both cases?

vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
...
+				v, err := newenv.get(ident.Lit)
+				if err != nil {
+					return nilValue, newError(ee, err)
+				}
+				letsStmts = append(letsStmts, e)
+				bodies = append(bodies, caseStmt.Stmts)
+				cases = append(cases, reflect.SelectCase{
+					Dir:  reflect.SelectRecv,
+					Chan: v,
+					Send: reflect.ValueOf(nil),
```
MichaelS11: Changed `reflect.ValueOf(nil)` to `zeroValue` for all?

vm/vmOperators_test.go
```diff
@@ -477,6 +477,42 @@ func TestIf(t *testing.T) {
...
+
+		// test 1 channel
+		{Script: `a = make(chan int64, 1); a <- 1; select {case <-a: return 1}`, RunOutput: int64(1)},
+
+		// test 2 channels
+		{Script: `a = make(chan int64, 1); b = make(chan int64, 1); a <- 1; select {case <-a: return 1; case <-b: return 2}`, RunOutput: int64(1)},
+		{Script: `a = make(chan int64, 1); b = make(chan int64, 1); b <- 1; select {case <-a: return 1; case <-b: return 2}`, RunOutput: int64(2)},
+
+		// test default
+		{Script: `a = make(chan int64, 1); b = make(chan int64, 1); select {case <-a: return 1; case <-b: return 2; default: return 3}`, RunOutput: int64(3)},
```
MichaelS11: Maybe add a default only case? Also a case that has default but default is not selected?

vm/vmOperators_test.go
```diff
@@ -477,6 +477,42 @@ func TestIf(t *testing.T) {
...
+func TestSelect(t *testing.T) {
+	os.Setenv("ANKO_DEBUG", "1")
+	tests := []testlib.Test{
+		// test run errors
+		{Script: `select {case a = <-b: return 1}`, RunError: fmt.Errorf("undefined symbol 'b'"), RunOutput: nil},
+		{Script: `select {case b = 1: return 1}`, RunError: fmt.Errorf("invalid operation"), RunOutput: nil},
+		{Script: `select {case 1: return 1}`, RunError: fmt.Errorf("invalid operation"), RunOutput: nil},
+		{Script: `select {case <-a: return 1}`, RunError: fmt.Errorf("undefined symbol 'a'"), RunOutput: nil},
+		{Script: `select {case if true { }: return 1}`, RunError: fmt.Errorf("invalid operation"), RunOutput: nil},
+		{Script: `a = make(chan int64, 1); a <- 1; select {case b.c = <-a: return 1}`, RunError: fmt.Errorf("undefined symbol 'b'"), RunOutput: nil},
```
MichaelS11: Add a scope check test? Make sure case assignment does go outside of select if variable not created ahead of time.

vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
...
+		for _, selectCaseStmt := range body.Cases {
+			caseStmt := selectCaseStmt.(*ast.SelectCaseStmt)
+			switch e := caseStmt.Expr.(type) {
+			case *ast.LetsStmt:
+				ee, ok := e.Rhss[0].(*ast.ChanExpr)
+				if !ok {
+					return nilValue, newStringError(e.Rhss[0], "invalid operation")
+				}
+				ident, _ := ee.Rhs.(*ast.IdentExpr)
+				v, err := newenv.get(ident.Lit)
```
MichaelS11: Add `v` and `err` near stop and reuse variables?

vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
...
+		if chosen == 0 {
+			return nilValue, ErrInterrupt
+		}
+		if letStmt := letsStmts[chosen]; letStmt != nil {
+			if _, err := invokeLetExpr(letStmt.Lhss[0], rv, newenv, ctx); err != nil {
+				return nilValue, newError(letStmt.Lhss[0], err)
+			}
+		}
+		if statements := bodies[chosen]; statements != nil {
+			var err error
```
MichaelS11: Create `err` near top?

vm/vmStmt.go
```diff
@@ -488,6 +488,85 @@ func runSingleStmt(stmt ast.Stmt, env *Env, ctx context.Context) (reflect.Value,
...
+				Chan: reflect.ValueOf(nil),
+				Send: reflect.ValueOf(nil),
+			})
+		}
+		chosen, rv, _ := reflect.Select(cases)
+		if chosen == 0 {
+			return nilValue, ErrInterrupt
+		}
+		if letStmt := letsStmts[chosen]; letStmt != nil {
+			if _, err := invokeLetExpr(letStmt.Lhss[0], rv, newenv, ctx); err != nil {
```
MichaelS11: Create `err` near top? Move err check to single line.

MichaelS11:
@alaingilbert 
I think this is a great addition, just would like to get a few notes addressed. Were you still going to work on this?

alaingilbert:
I don't have a lot of time to work on it.
I think there was a bug with
```
{Script: `select {default:}; return 1`, RunOutput: int64(1)},
```
which hang forever.
I'll try to figure it out.

alaingilbert:
I fixed the PR, everything is working well now.

MichaelS11:
@alaingilbert Could you look again at some of the previous comments?

vm/vmStmt.go
alaingilbert: I don't see a good way of doing this without the letsStmts and bodies.

vm/vmOperators_test.go
```diff
@@ -477,6 +477,52 @@ func TestIf(t *testing.T) {
 	testlib.Run(t, tests, nil)
 }
 
+func TestSelect(t *testing.T) {
+	os.Setenv("ANKO_DEBUG", "1")
+	tests := []testlib.Test{
+		// test parse errors
+		{Script: `select {default: return 6; default: return 7}`, ParseError: fmt.Errorf("multiple default statement"), RunOutput: int64(7)},
+		{Script: `a = make(chan int64, 1); a <- 1; select {case <-a: return 5; default: return 6; default: return 7}`, ParseError: fmt.Errorf("multiple default statement"), RunOutput: int64(5)},
+
```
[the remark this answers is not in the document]
alaingilbert: Do you know why the script seems to continue being executed even though there was a parsing error ?
Why do I get the RunError. Shouldn't it stop when it fails to parse ?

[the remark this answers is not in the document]
alaingilbert: I think the return should be outside the if condition here:
https://github.com/mattn/anko/blob/adf6aeefec2c5f68e58c142e1e5ae3be250289e9/internal/testlib/testlib.go#L44
So when it fails to parse, it always stop the execution.
What do you think ?

PR #289, opened 2019-01-17
strip comments before processing

tmclane:
This change strips comments from the source read from a file in non-interactive mode such that the comments do not cause the syntax check to fail.

### DIFF
--- a/anko.go
+++ b/anko.go
@@ -110,7 +110,9 @@ func main() {
 			}
 			code += string(b)
 		} else {
-			code = string(b)
+			for _, line := range strings.Split(code, "\n") {
+				code += strings.Split(line, "//")[0]
+			}
 		}
 
 		parser.EnableErrorVerbose()


MichaelS11:
I think comments should be handled by the parser, is it not working correctly? Do you have an example?

tmclane:
What I am seeing is this:

#### Anko Script (anko.ank)
```
x = 1
y = x + 1 / 2

// print using outside the script defined println function
```

#### Shell Check Run without Fix
```
~/go/src/github.com/mattn/anko$ ./anko -c anko.ank 
../../tmclane/ankotest/scripts/anko.ank:4:1 syntax error: unexpected '/
```

MichaelS11:
Thanks, will look into this more this week.

tmclane:
Shall I close this PR then?

MichaelS11:
I am unable to reproduce the issue.

I tried comments from interactive mode and they seemed to work fine.

I copied the "Anko Script Quick Start" from the readme into a file named anko.ank and ran it. It worked fine.

```
> anko.exe anko.ank
3
3
[1 2 3]
1
map[x:1]
2
3
4
```

Are you on the newest version of Anko?

tmclane:
Very strange.
Yes I git cloned it, checked out the `-c` branch and built the `anko` executable.
Bear in mind I am running on Linux. Could that be why I am seeing it and you can't reproduce it?

tmclane:
It's not the execution that fails it's the `syntax check` option. I did notice that branch is behind `master` at this time. I'll rebase and verify.

MichaelS11:
Tested on Linx, seems to be fine.

```
]$ ./anko anko.ank 
3
3
[1 2 3]
1
map[x:1]
2
3
4
```

The syntax check option was dropped so if there is a bug in that branch probably not going to get fixed.

tmclane:
I am closing this PR since it was specifically to fix an error in the `mattn:syntax-check` branch.

PR #290, opened 2019-01-22
Added ast operators

MichaelS11:
Helps make parser a little easier.

Next up is adding operator precedence testing.


MichaelS11:
@mattn Good to merge

mattn:
Thanks. Question, what is motivation of this?

MichaelS11:
Enable operator precedence is the main one. I think there are additional benefits like easier to read parser code and smaller switch statements.

mattn:
(I'm okay to merge this)

I wonder how the cost of function call vs switch case, or limit of stacks.

MichaelS11:
I wonder as well, we could probably do some benchmarks.

But really performance was not the goal. When I try to make changes to the parser my head spins. Really trying to split up the parser code to make it easier to manage.

mattn:
>When I try to make changes to the parser my head spins.

lol

Do you want to wait merge until taking benchmark? or merge?

MichaelS11:
:)

Merge please. 

I think the performance difference is small, just think it would be interesting to see.

mattn:
Thanks!

mattn:
FYI, anko is slowly than other implementation. (but I don't mind)

https://github.com/d5/tengo#benchmark

PR #297, opened 2019-02-02
Go Report Fixes and  Added TestUnknownCases

MichaelS11:
gofmt was used on a couple of files so the diff might shows a lot of changes but not really.


MichaelS11:
@mattn Good to merge

PR #299, opened 2019-02-07
Reorder parser, remove expr_op, change MapExpr to two slices

MichaelS11:


MichaelS11:
@mattn 
Side note:

With Go 1.7.x sometimes the value returned is the channel instead of the value returned from the channel. I have only seen this with 1.7.x so far.

	testlib.go:104: outputName b - received: []interface {}{1, (chan int64)(0xc420213a40), 3} - expected: []interface {}{1, 2, 3} - script: a = make(chan int64); go func() { a <- 1; a <- 2; a <- 3 }(); b = []; for i in a { b += i; if i > 2 { break } }
	testlib.go:105: received type: []interface {} - expected: []interface {}

I think it is a reflect select issue from this part of the code:

https://github.com/mattn/anko/blob/95dd1f3659fff7b85fcf32456e2d8dddd8be48b8/vm/vmStmt.go#L431-L435

MichaelS11:
@mattn Good to merge

mattn:
For the side note, is is a bug of anko?

MichaelS11:
I think it is a bug in the Go reflect Select code, but hard to tell. I have only seen it happen twice and only with 1.7.x so far.

It is nothing that was changed in this PR. It only started showing up after I added test code for the for chan loop. Since there was no test code for it before, no way to tell when the bug started.

mattn:
Thanks. BTY, do you have an example code that can repro?

MichaelS11:
It is this test:
```{Script: `a = make(chan int64); go func() { a <- 1; a <- 2; a <- 3 }(); b = []; for i in a { b += i; if i > 2 { break } }`, RunOutput: nil, Output: map[string]interface{}{"b": []interface{}{int64(1), int64(2), int64(3)}}},```

mattn:
Thank you

mattn:
It seems that this can be reproduced on Go master branch.

terminal7

MichaelS11:
If you do not get to it first, tomorrow I will see if I can get some simple Go code together (without Anko) that will reproduce it.

PR #300, opened 2019-02-07
Fix MemberExpr and Go Func. Added TestGoFunctionConcurrency

MichaelS11:


MichaelS11:
@mattn Good to merge

mattn:
Thank you

Issue #264, opened 2018-07-18
json: unsupported type: map[interface {}]interface {}


Chyroc:
new map data type is `map[interface {}]interface {}`

and `json.Marshal` of `"encoding/json"` not support it.

two options:
1. add buildin function dumps and loads
2. replace `"encoding/json"` to `"github.com/json-iterator/go"`, it support `interface{}` map key

or else better answer?

thank you .

Chyroc:
```go
package main

import (
	"fmt"
	"encoding/json"

	"github.com/json-iterator/go"
)

func main() {
	a, err := jsoniter.Marshal(map[interface{}]interface{}{"1": "1", 1: 2, false: 3,})
	fmt.Printf("%s - %v\n", a, err)

	b, err := json.Marshal(map[interface{}]interface{}{"1": "1", 1: 2, false: 3,})
	fmt.Printf("%s - %v\n", b, err)
}
```

output:
```
{"1":"1","1":2,"false":3} - <nil>
 - json: unsupported type: map[interface {}]interface {}
```

MichaelS11:
Could do something like this:

```
package main

import (
	"fmt"
	"log"

	"github.com/mattn/anko/packages"
	"github.com/mattn/anko/vm"
)

func main() {

	env := vm.NewEnv()
	packages.DefineImport(env)

	err := env.DefineType("mapStringInterface", map[string]interface{}{})
	if err != nil {
		log.Fatalf("DefineType error: %v\n", err)
	}

	script := `
fmt = import("fmt")
json = import("encoding/json")
a = make(mapStringInterface)
a["b"] = "b"
c, err = json.Marshal(a)
if err != nil {
	fmt.Println(err)
}
`

	_, err = env.Execute(script)
	if err != nil {
		log.Fatalf("execute error: %v\n", err)
	}

	dataInterface, err := env.Get("c")
	if err != nil {
		log.Fatalf("Get error: %v\n", err)
	}

	data := dataInterface.([]byte)
	fmt.Println(string(data))
}
```

MichaelS11:
@Chyroc 
Can this be closed?

Chyroc:
i think `json.Marshal` should support buildin type `map` without `mapStringInterface`

or we should add new function `map()` to new a map whitch is `mapStringInterface`

MichaelS11:
Just need better support for creating custom types inside of Anko. It is on my to do list. This way don't need to do the type outside of the script.

MichaelS11:
[...]

PR #304, opened 2019-02-13
Added create map type

MichaelS11:


MichaelS11:
@mattn Good to merge

mattn:
Should we add to the modification of README.md to clarify that anko treat map[string]interface{} and map[interface{}]interface {} separately?

MichaelS11:
@mattn How are they treated differently? If they are, I am not aware how they are treated differently.

```
a = map[interface]interface {"b":"b"}; a.b // b
a = map[string]interface {"b":"b"}; a.b // b
```

mattn:
Is this related on https://github.com/mattn/anko/issues/264, isn't ?

MichaelS11:
A little. Mostly it is for people to have basically the same syntax as Go. The only difference is that interface{} type is interface, which is not ideal but what works.

mattn:
Thanks for your explaining.

PR #313, opened 2019-06-27
Added string to byte & rune conversion

MichaelS11:


MichaelS11:
@mattn Good to go

mattn:
Thanks. If the string have two or more letters, this works without errors. hmm...I can't decide which is best yet.

MichaelS11:
Fix makeValue for struct ptr. Added strings.Builder.

@mattn Please merge

mattn:
>If the string have two or more letters, this works without errors. hmm...I can't decide which is best yet.

What do you think? If a string "foo" is passed for argument typed rune, should it work without error?

https://github.com/mattn/anko/pull/313/commits/ea4ba682721774d4ef573487e7be8a9ef096a505#diff-763921223b22a5e0fa21c13a3532ffe9R542

MichaelS11:
I am not sure which is best either. I am fine either way. It is a difficult choice.

mattn:
Could you please add code to make it error? If someone do not want it, let's remove the check.

mattn:
Thanks. I'll merge this after test passed.

MichaelS11:
Welcome :)

mattn:
Thank you

PR #325, opened 2020-03-02
Select statement

alaingilbert:
Add select to the language

examples:

```go
a = make(chan int64, 1)
select {
    case <-a:
        println("value received on a")
    default:
        println("default")
}
```
```go
a = make(chan int64, 1)
b = make(chan int64, 1)
b <- 2
val = 0
select {
    case val = <-a:
        println("value " + val + " received on a")
    case val = <-b:
        println("value " + val + " received on b")
}
```


parser/parser.go.y
```diff
@@ -379,6 +391,62 @@ stmt_for :
...
+	}
+
+stmt_select_default :
+	DEFAULT ':' compstmt
+	{
+		$$ = $3
+	}
+	| DEFAULT ':'
+	{
+		$$ = &ast.SelectCaseStmt{}
```
MichaelS11: Will adding SetPosition work?

ast/stmt.go
```diff
@@ -93,6 +93,26 @@ type ModuleStmt struct {
...
 
+// SelectStmt provide switch statement.
+type SelectStmt struct {
+	StmtImpl
+	Body Stmt
+}
+
+// SelectBodyStmt provide switch case statements and default statement.
+type SelectBodyStmt struct {
+	StmtImpl
```
MichaelS11: I do not think Body is needed.

vm/vmStmt.go
```diff
@@ -618,6 +618,99 @@ func (runInfo *runInfoStruct) runSingleStmt() {
...
 		runInfo.rv = nilValue
 
+	// SelectStmt
+	case *ast.SelectStmt:
+		e := runInfo.env
+		runInfo.env = e.NewEnv()
+
+		body := stmt.Body.(*ast.SelectBodyStmt)
+		letsExprs := []ast.Expr{nil}
+		bodies := []ast.Stmt{nil}
```
MichaelS11: Do not think these slices are needed.  Could a index be used to lookup in the Select Statment?

vm/vmStmt.go
```diff
@@ -618,6 +618,99 @@ func (runInfo *runInfoStruct) runSingleStmt() {
 		}
 		runInfo.rv = nilValue
 
+	// SelectStmt
+	case *ast.SelectStmt:
+		e := runInfo.env
+		runInfo.env = e.NewEnv()
```
MichaelS11: Should select be in a new env?

alaingilbert: Can you give me some information about what should and should not be in a new env ?
I inspired myself from the `SwitchStmt` which has a new env.
https://github.com/mattn/anko/blob/57d52a5bf01245ecdfd175f67e8bbfbe663c64a8/vm/vmStmt.go#L715-L717

MichaelS11: Each switch expression needed to be in a new scope (a new env) so that one switch expression did not affect another switch express. I am not sure if that is needed here.

PR #326, opened 2020-03-02
Move logic outside of main switch

alaingilbert:
I have a fork of this project.
One of the thing I did was to extract the logic out of the switch.
I find it easier to read and follow the code.
Are you interested by this change ?

### DIFF
--- a/vm/vmStmt.go
+++ b/vm/vmStmt.go
@@ -59,266 +59,363 @@ func (runInfo *runInfoStruct) runSingleStmt() {
 	}
 
 	switch stmt := runInfo.stmt.(type) {
-
-	// nil
 	case nil:
-
-	// StmtsStmt
 	case *ast.StmtsStmt:
-		for _, stmt := range stmt.Stmts {
-			switch stmt.(type) {
-			case *ast.BreakStmt:
-				runInfo.err = ErrBreak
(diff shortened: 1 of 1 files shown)


MichaelS11:
Not really seeing how it makes it easier to read or follow the code.

alaingilbert:
Well, having small functions that you can "goto definition" and go back.
Instead of a single function with a thousands lines...

I can see right away what are all the "statements" types that are handled by the `runSingleStmt` just by looking at the switch.
(instead of having to scroll through it, not knowing what is where.

MichaelS11:
Ok, I get what you are saying.

Note that all the statements are in https://github.com/mattn/anko/blob/master/ast/stmt.go already, so not that hard to see what statements are handled.

I am not partiularly for this change.

Plus I am not sure if the commit is clean. At least Github online diff is showing a lot of other diffrences.

alaingilbert:
The main point is code navigation & comprehension.

The git diff is indeed quite ugly
image

But the only thing I did was:
- take the logic in each `case`
- make a function out of it.

But yeah, no big deal, if you don't want this :)

MichaelS11:
Thanks for looking into this.

@mattn has the final say.

PR #337, opened 2020-05-15
Add go.mod

aisk:
This enables build anko on go module mode. Since anko have no any other dependencies, this change's only benifit is to build anko without GOPATH. And go module is the de facto depenency manage system I think we should use it for other project to embed anko easily.

### DIFF
--- a/go.mod
+++ b/go.mod
@@ -1,0 +1,3 @@
+module github.com/mattn/anko
+
+go 1.13


go.mod
```diff
@@ -0,0 +1,3 @@
+module github.com/mattn/anko
+
+go 1.14
```
mattn, line 3: Please fix to go 1.13

PR #347, opened 2021-03-21
add GetValueSymbols and GetTypeSymbols

issadarkthing:
I've been using anko for my [music player](https://github.com/issadarkthing/gomu) and I wanted to implement auto completion. By adding `GetValueSymbols` and `GetTypeSymbols`, I will be able to easily fetch symbols of the current scope. This will also help others with the same use case.


mattn:
LGTM, I'll look into it in later.

mattn:
Thank you

Issue #351, opened 2022-01-12
Methods definition


c4s4:
In Go, a method defined on a pointer like this:

```go
func (u *User) SayHello() {...}
```

May be called on a pointer or on an instance:

```go
user1 := &User{Name: "Michel"}
user1.SayHello()
user2 := User{Name: "Michel"}
user2.SayHello()
```

But this is not the case in Anko, as following example demonstrates:

```go
package main

import (
	"fmt"
	"log"

	"github.com/mattn/anko/env"
	"github.com/mattn/anko/vm"
)

type User struct {
	Name string
}

func (u *User) SayHello() {
	fmt.Printf("Hello %s!\n", u.Name)
}

func main() {
	e := env.NewEnv()
	user := User{Name: "Michel"}
	user.SayHello()
	err := e.Define("user", user)
	if err != nil {
		log.Fatalf("Define error: %v\n", err)
	}
	_, err = vm.Execute(e, nil, "user.SayHello()")
	if err != nil {
		log.Fatalf("Execute error: %v\n", err)
	}
}
```

Which prints:

```
$ go run main.go 
Hello Michel!
2022/01/12 18:42:13 Execute error: no member named 'SayHello' for struct
exit status 1
```

PR #354, opened 2023-06-05
Call pointer receiver method in struct passed by value.

ipsusila:
An attempt to fix issue #351


mattn:
Thank you. Could you please add test?

ipsusila:
> Thank you. Could you please add test?

Just add test cases with struct's instance passed to VM as pointer and value types.

mattn:
Thank you

Issue #358, opened 2023-10-13
Data race when running scripts concurrently


brutella:
When I run multiple scripts concurrently, I get a data race when the global variables [Packages](https://github.com/mattn/anko/blob/afcbc04aa5f5f6d3304b559484977aa2ebccb436/env/env.go#L32) and [PackageTypes](https://github.com/mattn/anko/blob/afcbc04aa5f5f6d3304b559484977aa2ebccb436/env/env.go#L36) are accessed [here](https://github.com/mattn/anko/blob/afcbc04aa5f5f6d3304b559484977aa2ebccb436/vm/vmExpr.go#L525) and [here](https://github.com/mattn/anko/blob/afcbc04aa5f5f6d3304b559484977aa2ebccb436/vm/vmExpr.go#L540).

I add and remove packages from `env.Packages` and `env.PackageTypes` before a script is executed. That's why I get a data race because the access to those global variables is not synchronised. I was able to fix this issue by allowing packages to be defined on an environment level – see pull request below.

Might be related to #34

mattn:
Thanks. Could you please add tests?

brutella:
Done

brutella:
@mattn Any chance to get this merged?

PR #359, opened 2023-10-13
Allow defining packages for environments

brutella:
This change enables environments to have a different set of packages. Accessing a package can now be protected by the environment mutex.

Fixes mattn/anko#358


brutella:
Any change to get this merged? cc @mattn

